PR #3, opened 2025-02-11
chore(APICLIENT-4): Initial backend setup (part 2)

brutusyhy:


Cargo.toml
```diff
@@ -1,10 +1,49 @@
...
 members = [
-    #
+    "crates/moss-cache",
+    "crates/moss-desktop",
+    "crates/moss-env",
+    "crates/moss-jsonlogic",
+    "crates/moss-jsonlogic-macro",
+    "crates/moss-tauri",
+    "crates/moss-text",
+    "tools/xtask",
```
g10z3r, line 11: We shouldn't move all the crates. Only the following should remain: moss-desktop, xtask

Makefile
```diff
@@ -1,24 +1,119 @@
...
+	@echo Cleaning Web Directory Cache...
+	@cd $(WEB_DIR) && $(PNPM) prune
+	@echo Cleaning Theme Generator Directory Cache...
+	@cd $(ADDON_THEME_DEFAULTS) && $(PNPM) prune
+	@echo Cleaning Icons Directory Cache...
+	@cd $(ICONS_DIR) && $(PNPM) prune
+	@echo Cleaning Desktop Models Directory Cache...
+	@cd $(DESKTOP_MODELS_DIR) && $(PNPM) prune
+	@echo Cleaning PNPM Store Cache...
+	$(PNPM) store prune
```
g10z3r, line 74: If we're moving this, then only the folders relevant to this project should remain here

crates/moss-desktop/src/menus.rs
```diff
@@ -0,0 +1,15 @@
...
+pub struct Menus {
+    state: DashMap<ReadOnlyStr, Vec<MenuItem>>,
+}
+
+impl Menus {
+    // TODO: rename to insert
+    pub fn append_menu_item(&mut self, key: ReadOnlyStr, item: MenuItem) {
+        self.state.entry(key).or_insert_with(Vec::new).push(item);
+    }
+}
```
g10z3r, line 15: This is not needed in this project

Makefile
brutusyhy: Oh my mistake!

g10z3r (approved):

PR #18, opened 2025-03-01
feat(SAPIC-48): Full integration of RW operations for collection and request handlers

brutusyhy:


crates/moss-collection/src/collection_manager.rs
```diff
@@ -104,7 +104,7 @@ impl CollectionManager {
     // collection descriptions should be sent to the frontend. This largely depends
     // on the library we choose to use for displaying hierarchical structures.
 
-    pub async fn index_collection(&self, path: PathBuf) -> Result<Arc<CollectionState>> {
+    pub(crate) async fn index_collection(&self, path: PathBuf) -> Result<Arc<CollectionState>> {
```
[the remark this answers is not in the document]
brutusyhy: Ok

crates/moss-collection/src/kdl/foundations/http.rs
```diff
@@ -1,8 +1,10 @@
 use crate::kdl::tokens::{HEADERS_LIT, PARAMS_LIT, URL_LIT};
-use kdl::{KdlDocument, KdlEntry, KdlNode, KdlValue};
+use kdl::{KdlDocument, KdlEntry, KdlNode};
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use ts_rs::TS;
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize, TS)]
```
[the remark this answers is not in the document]
brutusyhy: Looks like this is because we are using it in `CreateRequestInput`, which will be exported.

crates/moss-collection/src/kdl/foundations/http.rs
```diff
@@ -142,23 +168,35 @@ impl Into<KdlNode> for PathParamOptions {
     }
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct HeaderBody {
```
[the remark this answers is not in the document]
brutusyhy: Good idea!

crates/moss-collection/src/kdl/foundations/http.rs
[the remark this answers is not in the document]
g10z3r: Then this is incorrect. KDL should never be used in DTOs

crates/moss-collection/src/collection_handle.rs
```diff
@@ -54,28 +85,175 @@ impl CollectionState {
...
+        JsonValue::Array(_) => {
+            None
+            // TODO: Invalid type, logging
+        }
+        JsonValue::Object(_) => {
+            None
+            // TODO: Invalid type, logging
+        }
+    }
+}
```
g10z3r, line 152: As we discussed, the value of a **query parameter**, **header parameter**, and **path parameter** can only be a string. Therefore, it doesn't make sense to accept `JsonValue` if it can only be a string. Let's just accept a **string** directly.

crates/moss-collection/src/collection_handle.rs
```diff
@@ -54,28 +85,175 @@ impl CollectionState {
...
+        store: Arc<dyn CollectionRequestSubstore>,
+        state: CollectionState
+    ) -> Self {
+        Self {
+            fs,
+            store,
+            state: Arc::new(state),
+        }
+    }
+
```
g10z3r, line 256: Is this used anywhere?

crates/moss-collection/src/collection_handle.rs
```diff
@@ -85,139 +263,452 @@ impl CollectionHandle {
...
+                self.state.insert_request_handle(
+                    key,
+                    RequestHandle::new(
+                        self.fs.clone(),
+                        RequestState {
+                            name: name.clone(),
+                            order: None,
+                            typ: Some(method.clone().into()),
+                            // TODO: handling variants
+                            variants: Default::default(),
```
g10z3r, line 297: Let's extract this into a separate helper function. Essentially, we'll have one helper for handling HTTP types, and in the future, we can add another helper for handling other protocols

crates/moss-collection/src/collection_handle.rs
```diff
@@ -85,139 +263,452 @@ impl CollectionHandle {
...
+                        self.fs.clone(),
+                        RequestState {
+                            name: name.clone(),
+                            order: None,
+                            typ: None,
+                            variants: Default::default(),
+                        },
+                    ),
+                )?;
+                (String::new(), "get".to_string())
```
g10z3r, line 320: Let's define the `Default` trait for the protocol type structure and use its default value here. This way, we can manage the default protocol value from a single place.

crates/moss-collection/src/collection_manager.rs
```diff
@@ -157,37 +178,136 @@ impl CollectionManager {
...
+            (*write_lock).insert(new_path.clone(), handle);
+        }
+
 
         Ok(())
     }
+
+    pub async fn delete_collection(&self, path_buf: PathBuf) -> Result<()> {
+        let collections = self.collections().await?;
+        // FIXME: Is this checking necessary?
```
g10z3r, line 286: I think the functionality here should be as follows:

1. **Check if the folder exists**:  
   - If it **exists**, delete it and remove it from the map.  
   - If the collection is **not in the map**, this is an anomaly, but it doesn’t affect anything. Just log the anomaly and complete the operation successfully.

2. **If the folder has already been deleted from the file system but still exists in the map**:  
   - Log this anomaly.  
   - Respond with OK, since technically, nothing critical happened.
   
We will also continue monitoring all collection folders, so if a user manually deletes a collection, our watcher will detect the change in the background and clean up the collection map in memory.  

In theory, such anomalies should not occur but we will log all anomalies regardless.

crates/moss-collection/src/collection_handle.rs
brutusyhy: Good idea!

crates/moss-collection/src/collection_handle.rs
brutusyhy: I forgot to delete it lol

g10z3r (approved):

crates/moss-collection/src/collection_manager.rs
brutusyhy: Got it!

PR #30, opened 2025-03-02
feat(SAPIC-27): Frontend Implementation of collection sections

Enthceph:
added recursive Tree component with some DnD functionality
added IconGen from main project

run `pnpm install`
run `make gen-icons`


view/desktop/src/components/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,324 @@
...
+  node,
+  onNodeUpdate,
+  depth,
+  horizontalPadding,
+  nodeOffset,
+  treeId,
+  parentNode,
+}: TreeNodeComponentProps) => {
+  const paddingLeft = `${depth * nodeOffset + horizontalPadding}px`;
+  const paddingRight = `${horizontalPadding}px`;
```
g10z3r, line 24: You can use `useMemo` here to compute `paddingLeft` and `paddingRight` to avoid unnecessary recalculations on each render.

```tsx
  const paddingLeft = useMemo(
    () => `${depth * nodeOffset + horizontalPadding}px`,
    [depth, nodeOffset, horizontalPadding]
  );
  const paddingRight = useMemo(() => `${horizontalPadding}px`, [horizontalPadding]);
```

view/desktop/src/components/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const ulList = useRef<HTMLUListElement>(null);
+  const spanRef = useRef<HTMLSpanElement>(null);
```
g10z3r, line 28: I would slightly rename the variable names for element references to improve clarity (if these names are meaningful).

```tsx
  const buttonRef = useRef<HTMLButtonElement>(null);
  const childrenListRef = useRef<HTMLUListElement>(null);
  const dropAreaRef = useRef<HTMLSpanElement>(null);
```

view/desktop/src/components/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,324 @@
...
+  const [dropAllowance, setDropAllowance] = useState<boolean | null>(null);
+
+  const handleFolderClick = () => {
+    if (!node.isFolder) return;
+
+    onNodeUpdate({
+      ...node,
+      isExpanded: !node.isExpanded,
+    });
+  };
```
g10z3r, line 40: Use callbacks here as well, and try to group similar code together. For example, keep all callbacks in one place sequentially, and so on.

```tsx
  const [preview, setPreview] = useState<HTMLElement | null>(null);
  const [dropAllowance, setDropAllowance] = useState<boolean | null>(null);
  const [renaming, setRenaming] = useState(false);

  const handleFolderClick = useCallback(() => {
    if (!node.isFolder) return;
    onNodeUpdate({ ...node, isExpanded: !node.isExpanded });
  }, [node, onNodeUpdate]);

  const handleButtonKeyUp = useCallback((e: React.KeyboardEvent<HTMLButtonElement>) => {
    if (e.key === "F2" && document.activeElement === e.currentTarget) {
      setRenaming(true);
    }
  }, []);
  
    const handleFormSubmit = useCallback(
    (newId: string) => {
      onNodeUpdate({ ...node, id: newId });
      setRenaming(false);
    },
    [node, onNodeUpdate]
  );

  const handleFormCancel = useCallback(() => {
    setRenaming(false);
  }, []);
  
    const handleExpandAll = useCallback(() => {
    const newNode = expandAllNodes(node);
    onNodeUpdate({ ...node, childNodes: newNode.childNodes });
  }, [node, onNodeUpdate]);

  const handleCollapseAll = useCallback(() => {
    const newNode = collapseAllNodes(node);
    onNodeUpdate({ ...node, childNodes: newNode.childNodes });
  }, [node, onNodeUpdate]);
  ...
```

view/desktop/src/components/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,324 @@
...
+          </div>
+        </div>
+
+        <ul
+          ref={ulList}
+          className={cn({
+            "bg-green-600": dropAllowance === true,
+            "bg-red-600": dropAllowance === false,
+            "": dropAllowance === null,
+          })}
```
g10z3r, line 206: It's preferable to move this into a `useMemo` callback

```tsx
  const dropClass = useMemo(() => {
    if (dropAllowance === true) return "bg-green-600";
    if (dropAllowance === false) return "bg-red-600";
    return "";
  }, [dropAllowance]);
```

view/desktop/src/components/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+        if (!dropTarget || !sourceTarget) {
+          setDropAllowance(null);
+          return;
+        }
+
+        if (dropTarget?.node.uniqueId !== node.uniqueId) {
+          setDropAllowance(null);
+          return;
+        }
```
g10z3r, line 101: I think this can be combined

```tsx
        if (!dropTarget || !sourceTarget || dropTarget.node.uniqueId !== node.uniqueId) {
          setDropAllowance(null);
          return;
        }
```

view/desktop/src/components/Tree/useNodeRenaming.ts
g10z3r, line 1: I think callbacks can be introduced here as well for optimization

view/desktop/src/components/Tree/utils.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+export const findNodeByUniqueId = (tree: TreeNodeProps, uniqueId: string): TreeNodeProps | undefined => {
+    if (tree.uniqueId === uniqueId) {
+        return tree;
+    }
+
+    return tree.childNodes.find(child => findNodeByUniqueId(child, uniqueId));
+}
```
g10z3r, line 73: It looks like using `.find` here could lead to recursion.  

I would rewrite it like this:

```tsx
export const findNodeByUniqueId = (tree: TreeNodeProps, uniqueId: string): TreeNodeProps | undefined => {
  if (tree.uniqueId === uniqueId) {
    return tree;
  }

  for (const child of tree.childNodes) {
    const found = findNodeByUniqueId(child, uniqueId);
    if (found) return found;
  }
  return undefined;
};
```

view/desktop/src/components/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,332 @@
...
+        setDropAllowance(null);
+      },
+    });
+  }, [dropAllowance, node, treeId]);
+
+  const [renaming, setRenaming] = useState(false);
+  const handleButtonKeyUp = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+    if (e.key === "F2" && document.activeElement === e.currentTarget) {
+      setRenaming(true);
```
g10z3r, line 136: There should be a more advanced and cross-platform solution here since this won't work on macOS

view/desktop/src/components/Tree/NodeRenamingForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+  currentName: string | number;
+}
+
+export const NodeRenamingForm = ({ onSubmit, onCancel, restrictedNames, currentName }: NodeRenamingFormProps) => {
+  const ref = useRef<HTMLInputElement>(null);
+  const [value, setValue] = useState(String(currentName));
+
+  const handleInputKeyUp = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Escape") onCancel();
+  };
```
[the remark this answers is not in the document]
Enthceph: I don't think we need to wrap functions in useCallback.

1) useCallback doesn't prevent functions recreations, it only caches the function and gives you a cached function if needed.

There is a really useful note on react.dev about it.

https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere

изображение

2) the use of useCallback here is unnecessary because these functions are only used directly as event handlers within the component. They aren’t being passed to memorized child components or used in dependency arrays where stable references matter. There is no performance benefit to wrapping these functions in useCallback, so it’s simpler to define them inline.

Not to mention that the NodeRenamingForm is not a heavy component and will appear only from time to time. Trying to optimize it feels like we are fixing performance problems that aren't there.

view/desktop/src/components/Tree/NodeRenamingForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <input
+        ref={ref}
+        value={value}
+        onChange={(e) => setValue(e.target.value)}
+        autoFocus
+        minLength={1}
+        maxLength={100}
+        className="flex gap-1 w-full min-w-0 grow items-center cursor-pointer focus-within:outline-none relative"
+        onKeyUp={handleInputKeyUp}
+        onBlur={restrictedNames.includes(String(value)) ? onCancel : handleSubmit}
```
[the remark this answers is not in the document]
Enthceph: I did extract it into it's own function but wrapping it into useCallback is useless. 
1) this callback depends on input value, meaning it will cache itself on every input change. 
2) it is dependent on all the props the component itself has

view/desktop/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,82 @@
...
+          setTree((prevTree) => {
+            const updatedTree = addNodeToFolder(prevTree, target.node.uniqueId, source.node);
+            return sortNode(updatedTree);
+          });
+        }
+        if (source.treeId === treeId) {
+          setTree((prevTree) => removeNodeFromTree(prevTree, source.node.uniqueId));
+        }
+      }
+    };
```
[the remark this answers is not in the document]
Enthceph: 1) useCallback doesn't prevent function recreactions
2) the handleMoveTreeNode function is created inside useEffect that is executed only once on component mount(and it has a treeId as a dependency, but treeId never changes). Extracting it into its own function will not change the amount of function recreations

view/desktop/src/components/Tree/TreeNode.tsx
Enthceph: In Tree node component there are 7 dependencies(node,onNodeUpdate,depth,horizontalPadding,nodeOffset,treeId,parentNode). But the only props that actially change are node and onNodeUpdate. As again wrapping theese functions in useCallback that is dependent on the only props that change will achive no real benefit

view/desktop/src/components/Tree/TreeNode.tsx
Enthceph: Memorizing if/else statement seems reductionist. Adding a check for cache before checking for value that could have only 3 states doesn't achieve much optimization.

Not to mention dropAllowance is the cause if the rerenders, because the state of dropAllowance paints node into red/green colors depending on it's state.

Funnily enough there is a way to prevent rerenders, but it will require working with node's DOM directly.

изображение

view/desktop/src/components/Tree/utils.ts
g10z3r: Yes, recursion is indeed needed here, but the problem isnt the lack of recursion, its how the `.find` method is being used. I might be wrong, but I believe that when we write:

```tsx
tree.childNodes.find(child => findNodeByUniqueId(child, uniqueId));
```

the callback function returns the found node (or `undefined`), and if the found node is **truthy**, `.find` will return the **immediate child** for which the recursive function call returned a non-null value. This means we might **lose the result of deep recursion**, instead of returning the actual found node, it returns the child element whose call produced a **truthy** result.

In contrast, when using a `for-of` loop, we explicitly check the result of the recursive call and immediately return the found value, ensuring the function behaves correctly. Does that make sense?

view/desktop/src/components/Tree/TreeNode.tsx
g10z3r: Yes, just remove this part if it hasn't been removed already

g10z3r (approved):

jevgenis33 (approved):

PR #31, opened 2025-03-02
feat(SAPIC-55): Added temporary custom Head Bar with controls

jevgenis33:


g10z3r:
Keep only the gray strip at the top, where a tag will be added to allow dragging the window. Components from the previous project should not be here.. we won’t be using them

g10z3r (approved):

PR #37, opened 2025-03-04
feat(SAPIC-60): [Backend] Basic logging management functionality

brutusyhy:
Implement a `LoggingService` that sets up tracing, tracing subscriber and tracing appender for producing structured logs in a JSON format to files. Also implement a basic filter functionality for querying the logs.


crates/moss-logging/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+use tracing_appender::non_blocking::WorkerGuard;
+use tracing_appender::rolling::Rotation;
+use tracing_subscriber::filter::filter_fn;
+use tracing_subscriber::fmt::format::{FmtSpan, JsonFields};
+use tracing_subscriber::fmt::time::ChronoLocal;
+use tracing_subscriber::prelude::*;
+
+use moss_session::SessionService;
+
+type LogEntry = JSONValue;
```
g10z3r, line 26: I think this is unnecessary, better to keep type transparency here.

crates/moss-logging/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+use serde_json::Value as JSONValue;
+use std::collections::HashSet;
+use std::ffi::OsStr;
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+use std::sync::Arc;
+use std::{fs, io};
+#[allow(unused_imports)]
```
g10z3r, line 15: Do we need imports here that are not used?

crates/moss-logging/src/lib.rs
```diff
@@ -0,0 +1,303 @@
+mod tokens;
+
+use crate::tokens::*;
+use anyhow::Result;
+use chrono::{NaiveDate, Utc};
+use serde_json::Value as JSONValue;
```
g10z3r, line 6: We need to settle on a consistent alias name.

It looks like we use JsonValue in other parts of the code, so we should stick to one alias format and use it consistently.

crates/moss-logging/src/tokens.rs
g10z3r, line 1: What are these token literals? We don’t use them anywhere in KDL, do we?

crates/moss-session/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "moss_session"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+uuid = { version = "1.15.1", features = ["v4"] }
```
g10z3r, line 7: probably it would be better to transfer it to the workspace level, otherwise there will be errors during the audit 😅

crates/moss-session/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+pub struct SessionService {
+    uuid: String,
+}
+
+impl SessionService {
+    pub fn init() -> SessionService {
```
g10z3r, line 6: If this is a constructor function, its name should be new, and it should return Self

crates/moss-session/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+        SessionService {
+            uuid: uuid::Uuid::new_v4().to_string(),
```
g10z3r, line 8: I don't think it's necessary to cast it to a string right away. The type casting should happen when required

crates/moss-logging/src/lib.rs
brutusyhy: Apparently they are used, so it's a false warning. I tried removing them but then some methods can't be used.

crates/moss-logging/src/lib.rs
brutusyhy: Ok

crates/moss-logging/src/lib.rs
brutusyhy: My bad! I hadn't checked the casing in other places lol

crates/moss-logging/src/tokens.rs
brutusyhy: Actually it shouldn't be labeled tokens, it's just constants. Maybe I should move it to the main file.

crates/moss-session/Cargo.toml
brutusyhy: That's right!

crates/moss-session/src/lib.rs
brutusyhy: I see

view/desktop/bin/src/lib.rs
```diff
@@ -85,6 +65,12 @@ pub fn run() {
                 sled::open("../../../sleddb").expect("failed to open a connection to the database");
             let sled_manager = SledManager::new(db).expect("failed to create the Sled manager");
```
g10z3r, line 67: All services should be created in a lazy callback like all other services do

g10z3r (approved):

PR #44, opened 2025-03-07
feature(SAPIC-62): [Backend] App scope logs

brutusyhy:
Introduce AppScope logs and provide methods for `LoggingService` for generating structured logging


crates/moss-logging/src/lib.rs
```diff
@@ -199,21 +179,257 @@ impl LoggingService {
...
+                )
+            }
+        }
+    }
+
+    pub fn debug(&self, scope: LogScope, payload: LogPayload) {
+        match scope {
+            LogScope::App => {
+                debug!(
+                    target: "app",
```
g10z3r, line 354: I think these string representations of scope should be defined as constants at the global level

view/desktop/bin/src/lib.rs
```diff
@@ -71,8 +71,11 @@ pub fn run() {
             let session_service = SessionService::new();
             // FIXME: In the future, we will place logs at appropriate locations
             // Now we put `logs` folder at the project root for easier development
-            let logging_service =
-                LoggingService::new(Path::new("../../../logs"), &session_service)?;
+            let logging_service = LoggingService::new(
+                Path::new("../../../logs/app"),
```
g10z3r, line 75: Use the path to the logs by setting it through the Makefile exported variable

crates/moss-logging/src/lib.rs
brutusyhy: Good idea

view/desktop/bin/src/lib.rs
brutusyhy: Ok

crates/moss-logging/src/lib.rs
```diff
@@ -199,21 +166,247 @@ impl LoggingService {
...
+
+impl LoggingService {
+    pub fn new(
+        app_log_path: &Path,
+        session_log_path: &Path,
+        session_service: &SessionService,
+    ) -> Result<LoggingService> {
+        let standard_log_format = tracing_subscriber::fmt::format()
+            .with_file(false)
+            .with_line_number(false)
```
g10z3r, line 217: Lets use iterators to traverse the elements of the vectors, allowing us to move them instead of cloning

Something like: 

```rust 
fn merge_logs_chronologically(
    a: Vec<(DateTime<FixedOffset>, JsonValue)>,
    b: Vec<(DateTime<FixedOffset>, JsonValue)>,
) -> Vec<(DateTime<FixedOffset>, JsonValue)> {
    let mut iter_a = a.into_iter();
    let mut iter_b = b.into_iter();
    let mut merged = Vec::with_capacity(iter_a.size_hint().0 + iter_b.size_hint().0);

    let mut next_a = iter_a.next();
    let mut next_b = iter_b.next();

    while let (Some(ref a_val), Some(ref b_val)) = (next_a.as_ref(), next_b.as_ref()) {
        if a_val.0 <= b_val.0 {
            merged.push(next_a.take().unwrap());
            next_a = iter_a.next();
        } else {
            merged.push(next_b.take().unwrap());
            next_b = iter_b.next();
        }
    }

    if let Some(val) = next_a {
        merged.push(val);
        merged.extend(iter_a);
    }
    if let Some(val) = next_b {
        merged.push(val);
        merged.extend(iter_b);
    }

    merged
}
```

crates/moss-logging/src/models/primitives.rs
g10z3r, line 1: These types seem redundant as their names are a bit misleading, I would expect it to be a Path type. Instead of creating an alias, it is better to rename the fields in the structure for clarity if more clarity is needed

brutusyhy: I see!

PR #46, opened 2025-03-07
devops(SAPIC-51): Write a script for automatic exporting of bindings from the index file

brutusyhy:
Create a `ts_exports_injector.py` script to recursively walk through the `bindings` folders in crates with TS exports, and generate the export statements in their `index.ts`. Also automatically formats the generated bindings with `pnpm format`


misc/ts_exports_injector.py
```diff
@@ -0,0 +1,9 @@
+import os
+from pathlib import Path
+
+if __name__ == "__main__":
+    with open("index.ts", mode="w") as f:
+        for root, dirs, files in os.walk("bindings"):
+            for file in files:
+                import_path = "./" + root.replace("\\", "/") + "/" + Path(file).stem
+                f.write(f"export * from \"{import_path}\";\n")
```
g10z3r, line 9: Buffer everything that will be written to the file in memory and perform a single open-and-write operation.

Otherwise, with a large number of files and crates with models, this will become a slow operation.

brutusyhy: Ok!

Makefile
```diff
@@ -57,6 +57,7 @@ define gen_models
 gen-$(1)-models:
 	@$(CARGO) test export_bindings_ --manifest-path $($(2))/Cargo.toml
 	@$(CARGO) build --manifest-path $($(2))/Cargo.toml
+	@cd $($(2)) && @$(PNPM) format
```
g10z3r, line 60: Here, we need to either:
	1.	Remove the @ symbol before calling pnpm, or
	2.	Remove the formatting command from this part since it won’t work on Linux/macOS.

But in theory, it should work on Windows without @, since we don’t use it in other places when calling pnpm.

g10z3r (approved):

Makefile
brutusyhy: Let me give it a try!

PR #50, opened 2025-03-11
devops(SAPIC-76): Constant formatting confusion

brutusyhy:
Disable bindings-related scripts in the `build.rs` to prevent any unnecessary rebuild triggered by modifying the files in the build process, and call `ts_imports_injector.py` and `ts_exports_injector.py` directly in the `make gen-models` script.


Makefile
```diff
@@ -23,12 +23,13 @@ DESKTOP_DIR := view/desktop
...
 XTASK_DIR := tools/xtask
-
+TS_IMPORT_INJECTOR := misc/ts_imports_injector.py
+TS_EXPORT_INJECTOR := misc/ts_exports_injector.py
 # --- Executables ---
 PNPM := pnpm
 CARGO := cargo
 RUSTUP := rustup
-
+PYTHON := python
```
g10z3r, line 32: This won't work on linux/macos

crates/moss-collection/build.rs
```diff
@@ -1,29 +1,29 @@
...
-                status
-                    .code()
-                    .map_or(ExitCode::FAILURE, |code| ExitCode::from(code as u8))
-            }
-        })
-        .unwrap_or_else(|err| {
-            eprintln!("Failed to execute command: {}", err);
-            ExitCode::FAILURE
-        })
+    // #[cfg(target_os = "windows")]
```
g10z3r, line 4: We need to fix the auto-formatting issue, not just move it elsewhere

g10z3r, line 4: We need to fix the auto-formatting issue, not just move it elsewhere

Makefile
brutusyhy: I see.

crates/moss-collection/build.rs
brutusyhy: But there will always be a delay between the generation of bindings and formatting, and I assume tauri will pick that up even if we call a formatting here.

g10z3r: Okay, if we’re removing everything from this file, then there’s no point in keeping the file

brutusyhy: Then let's just remove them I think. I also couldn't execute pnpm format here for some reason.

Makefile
```diff
@@ -56,7 +63,8 @@ define gen_models
 .PHONY: gen-$(1)-model
 gen-$(1)-models:
 	@$(CARGO) test export_bindings_ --manifest-path $($(2))/Cargo.toml
-	@$(CARGO) build --manifest-path $($(2))/Cargo.toml
+	@cd $($(2)) && $(PYTHON) ../../$(TS_IMPORT_INJECTOR) package.json
```
g10z3r, line 66: Can we use the current working directory here?

brutusyhy: Maybe? Let me try it out

Enthceph (approved):

PR #51, opened 2025-03-11
feat(SAPIC-54): Implementation of additional functionality for the collection tree

Enthceph:
added:

- collection creation from folders and files with drag and drop 
- collection renaming
- collection order change with drag and drop

Fixed unnecessary scroll appearance


view/desktop/src/components/SidebarExplorer.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+import "@repo/moss-tabs/assets/styles.css";
+
+import { cn, swapListById } from "@/utils";
+import { Edge } from "@atlaskit/pragmatic-drag-and-drop-hitbox/dist/types/types";
+
+import { Collection, CreateNewCollectionFromTreeNodeEvent } from "./Tree/types";
+import { getActualDropSourceTarget } from "./Tree/utils";
+
+export const SidebarExplorer = () => {
```
g10z3r, line 18: We will have multiple explorers within the sidebar, so components should have more descriptive names.  In general, views within the sidebar are referred to as `TreeView`, based on our terminology.  So, the most appropriate name for the collection view would be `CollectionTreeView`.

g10z3r (approved):

PR #57, opened 2025-03-14
feat(SAPIC-94): [Backend] Implement support for the remaining RAW body types

brutusyhy:
Implement support for the following RAW body types with unit tests:
Text, JSON, HTML, XML

Right now, an error will be thrown if the last three types contains invalid content when parsing. An exception is HTML, where you can ignore the validation errors if you enabled relaxed mode for html parsing, since even technically malformed HTML can often be rendered by the browser (or accepted by the server).

A remaining issue is that an error will be thrown if the html being parsed does not begin with a `<!DOCTYPE>` declaration. It can be ignored when the relaxed parsing mode is enabled, but ideally this error should be separated from other html syntax errors, and I still need to figure out the best way to achieve it.

### DIFF
--- a/crates/moss-collection/src/collection_handle.rs
+++ b/crates/moss-collection/src/collection_handle.rs
@@ -1,5 +1,5 @@
 use crate::models::collection::RequestType;
-use crate::models::types::request_types::RequestBody;
+use crate::models::types::request_types::{RawBodyType, RequestBody};
 use crate::{
     kdl::foundations::http::{
         HeaderParamBody, HeaderParamOptions, HttpRequestFile, PathParamBody, PathParamOptions,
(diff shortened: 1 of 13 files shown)


crates/moss-collection/src/collection_handle.rs
```diff
@@ -200,7 +200,20 @@ fn create_http_requestfile(
...
+                RawBodyType::Json(raw) => {
+                    crate::kdl::body::RequestBody::Raw(crate::kdl::body::RawBodyType::Json(raw))
+                }
+                RawBodyType::Html(raw) => {
+                    crate::kdl::body::RequestBody::Raw(crate::kdl::body::RawBodyType::Html(raw))
+                }
+                RawBodyType::Xml(raw) => {
+                    crate::kdl::body::RequestBody::Raw(crate::kdl::body::RawBodyType::Xml(raw))
+                }
+            },
```
g10z3r, line 216: Let's move the `RequestBody::Raw` branch with the `match` statement into a separate function

crates/moss-collection/src/collection_manager.rs
g10z3r, line 1: This file should not be modified as part of this ticket. Additionally, in another branch, I am deleting the `CollectionManager` entity

crates/moss-collection/src/storage/collection_store.rs
g10z3r, line 1: This file should not be modified as part of this ticket. Same situation, there are many changes to this file in a parallel PR.

crates/moss-collection/src/collection_manager.rs
brutusyhy: Sorry I just ran a cargo fmt lol, I'll fix it

g10z3r (approved):

PR #60, opened 2025-03-18
feat(SAPIC-95): [Backend] Implement support for the remaining body types

brutusyhy:
Form-data example:
body type=form-data {
    key1 {
        type "text"
        value "value"
    }
    key2 {
        type "file"
        value "path/to/file"
    }
}

x-www-form-urlencoded example:
body type=urlencoded {
    key1 {
        value "value"
    }
    key2 {
        value "value"
    }
}

Binary example:
body type=binary path="path/to/file"


crates/moss-collection/src/kdl/foundations/body.rs
```diff
@@ -72,14 +288,99 @@ fn prepare_raw_body_node(node: &mut KdlNode, raw_body: RawBodyType) {
...
+        println!("{}", node.to_string());
+    }
+
+    #[test]
+    fn test_writing_binary_body() {
+        let request_body = RequestBody::Binary(PathBuf::from("path/to/file"));
+        let node: KdlNode = request_body.into();
+        println!("{}", node.to_string());
+    }
+}
```
g10z3r, line 386: These aren’t exactly tests; they’re more like visual tests. We need to create something like expected data and compare it with the actual results

brutusyhy: Indeed, I will fix these parts

crates/moss-collection/src/collection_handle.rs
```diff
@@ -201,26 +202,71 @@ fn create_http_requestfile(
...
+                propagate: item.options.propagate,
+            }
+        });
+    }
+    crate::kdl::body::RequestBody::UrlEncoded(map)
+}
+
+fn map_binary_to_kdl(path: &str) -> crate::kdl::body::RequestBody {
+    crate::kdl::body::RequestBody::Binary(PathBuf::from(path))
+}
```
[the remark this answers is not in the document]
brutusyhy: But we don't have our own type for the other three, so we can't implement From trait directly on them I assume?

g10z3r (approved):

PR #62, opened 2025-03-19
feat(SAPIC-104): [Backend] Finalization and refactoring of the collection crate

brutusyhy:
1. Implement the request deletion method for the Collection type.
2. Remove SledDB from the project and deleted all related code.
3. Remove the external crate mockall and deleted all associated code.
4. Remove all test placeholder tests and enabled tests in the CI pipeline.


crates/moss-collection/src/collection.rs
```diff
@@ -371,10 +388,53 @@ impl Collection {
...
+
+        let request_store = self.state_db_manager.request_store();
+        let (mut txn, table) = request_store.begin_write()?;
+        table.remove(
+            &mut txn,
+            request_dir_relative_path.to_string_lossy().to_string(),
+        )?;
+
+        std::mem::drop(lease_request_data);
+        requests_lock.remove(request_key)?;
```
g10z3r, line 434: Here, you need first check if the slot is leased. If it is not leased, then immediately remove the slot from the map and drop the _mut_ mutex. This ensures that the data is hidden from the frontend, and regardless of how subsequent operations complete, the user won’t see an error, only logs will reflect any failure.  

Next, if we fail to physically delete the data from the FS, we will handle it later in the background during the self-healing process

crates/moss-collection/src/collection.rs
```diff
@@ -520,6 +583,50 @@ mod tests {
...
+        let create_request_output = collection
+            .create_request(CreateRequestInput {
+                name: request_name.clone(),
+                relative_path: None,
+                payload: None,
+                url: None,
+            })
+            .await
+            .unwrap();
+
```
g10z3r, line 615: After the creation operation, I think we should check that the request **actually exists** in the request map to be sure that we're deleting something we just created, and then proceed with the deletion.

crates/moss-collection/src/collection.rs
```diff
@@ -200,7 +202,7 @@ impl Collection {
                     self.known_requests_paths.insert(request_dir_path.clone());
                     requests.insert(CollectionRequestData {
                         name: indexed_request_entry.name,
-                        request_dir_path,
+                        request_dir_relative_path: request_dir_path,
```
g10z3r, line 205: Should we rename the variable `request_dir_path` if it actually represents `request_dir_relative_path`?

crates/moss-db/src/bincode_table.rs
```diff
@@ -119,12 +119,13 @@ where
...
-mod test {
+#[cfg(test)]
+mod tests {
     use super::*;
     use crate::{DatabaseClient, ReDbClient};
 
     #[test]
-    fn test_scan() {
+    fn scan() {
         let client: ReDbClient = ReDbClient::new("test_scan.db").unwrap();
```
g10z3r, line 129: Maybe we should also do this in the **tests** folder and generate the database name with a random string in it, so that parallel tests don’t compete within the same database?

crates/moss-collection/src/storage.rs
```diff
@@ -1,18 +1,14 @@
...
-pub use collection_request_substore::SledCollectionRequestSubstore;
-
 use anyhow::Result;
 use moss_db::{bincode_table::BincodeTable, Transaction};
 use std::{collections::HashMap, path::PathBuf, sync::Arc};
 
 use crate::models::storage::{CollectionEntity, RequestEntity};
 
-#[cfg_attr(test, mockall::automock)]
 pub trait CollectionRequestSubstore: Send + Sync + 'static {}
```
g10z3r, line 12: I think we should be able tщ remove the entire `CollectionRequestSubstore` trait now

sleddb/conf
g10z3r, line 1: We should remove all files related to this database in this PR

view/sleddb/conf
g10z3r, line 1: [Same](https://github.com/moss-foundation/sapic/pull/62#discussion_r2003938598) here

crates/moss-collection/src/collection.rs
brutusyhy: How should we handle when it's leased?

crates/moss-db/src/bincode_table.rs
brutusyhy: Good idea!

crates/moss-collection/src/collection.rs
```diff
@@ -99,10 +102,28 @@ impl Collection {
...
+    }
+
+    pub fn reload_db(&mut self) -> Result<()> {
+        let state_db_manager_impl = StateDbManagerImpl::new(&self.path).context(format!(
+            "Failed to open the collection {} state database",
+            self.path.display()
+        ))?;
+        self.state_db_manager = Some(Arc::new(state_db_manager_impl));
+        Ok(())
+    }
```
g10z3r, line 126: I don’t think this needs to be extracted into a separate function in this case. The database drop operation should only be executed during collection reset.

Keeping everything in one function ensures that this operation cannot be executed elsewhere, preventing unintended behavior.

PR #64, opened 2025-03-19
refactor(SAPIC-99): standardization of the table initialization approach

g10z3r:
@brutusyhy


brutusyhy:
Nice! I'll have a look at it when I'm back home!

PR #65, opened 2025-03-19
refactor(SAPIC-112): implementation of test cases for core collections and workspaces

g10z3r:


crates/moss-workspace/src/workspace_manager.rs
```diff
@@ -15,236 +18,336 @@ use crate::{
...
+    global_db_manager: Option<Arc<dyn GlobalDbManager>>,
+    current_workspace: ArcSwapOption<(WorkspaceKey, Workspace)>,
+    known_workspaces: OnceCell<RwLock<WorkspaceInfoMap>>,
 }
 
 impl WorkspaceManager {
-    pub fn new(fs: Arc<dyn FileSystem>, workspaces_dir: PathBuf) -> Self {
-        Self {
+    pub fn new(fs: Arc<dyn FileSystem>, workspaces_dir: PathBuf) -> Result<Self> {
+        let global_db_manager = GlobalDbManagerImpl::new(&workspaces_dir).context("Failed to open the global state database")?;
```
g10z3r, line 77: `WorkspaceManager` does not own the global state database, so it should be passed into it as a parameter. The global state DB will be used in many parts of the application.

crates/moss-workspace/src/workspace_manager.rs
```diff
@@ -15,236 +18,336 @@ use crate::{
...
+
+    #[error("unknown error: {0}")]
+    Unknown(#[from] anyhow::Error),
+}
+
+type WorkspaceInfoMap = LeasedSlotMap<WorkspaceKey, WorkspaceInfo>;
 
 pub struct WorkspaceManager {
     fs: Arc<dyn FileSystem>,
+    // TODO: Should we allow creating workspaces at arbitrary locations?
```
g10z3r, line 68: `WorkspaceManager` does not create the directory where all workspaces will be stored—this directory will be created during the application installation process.

crates/moss-collection/src/sanitizer.rs
g10z3r, line 1: Let's move this into the `fs` crate since it will be used across multiple crates. Also, I think this applies not just to directory names but to file names as well, so the function name should be more general.

crates/moss-workspace/Cargo.toml
```diff
@@ -17,6 +17,8 @@ ts-rs.workspace = true
 tokio.workspace = true
 thiserror.workspace = true
 slotmap.workspace = true
+validator = { workspace = true, features = ["derive"] }
+regex = "1.11.1"
```
g10z3r, line 21: The dependency should be moved up to the workspace level

crates/moss-workspace/src/workspace_manager.rs
```diff
@@ -15,236 +18,336 @@ use crate::{
...
 
-    pub fn open_workspace(&self, input: OpenWorkspaceInput) -> Result<()> {
-        let workspace = Workspace::new(input.path, self.fs.clone())?;
-        self.current_workspace.store(Some(Arc::new(workspace)));
-        Ok(())
-    }
+        txn.commit()?;
+
+        Ok(CreateWorkspaceOutput {
+            key: workspace_key.as_u64()
```
g10z3r, line 185: There's no point in using a transaction here. Just create a `create()` method on the store, and in the function, first create the directory, if it's successfully created, then save the data to the DB.

PR #68, opened 2025-03-22
feat(SAPIC-88): Implemented custom scrollbar to moss-tabs library

jevgenis33:
image


sleddb/conf
g10z3r, line 1: You might need to pull from main, this folder was deleted because we no longer use SledDB. If pulling doesn’t resolve it, you can simply delete the folder both locally and from your branch

g10z3r (approved):

PR #75, opened 2025-03-28
feat(SAPIC-114): Frontend Breadcrumbs

Enthceph:


view/desktop/src/store/collections/index.ts
```diff
@@ -0,0 +1,23 @@
+import { create } from "zustand";
+
+import { Collection } from "@/components/Tree/types";
+
+import AzureDevOpsTestCollection from "../../assets/AzureDevOpsTestCollection.json";
+import SapicTestCollection from "../../assets/SapicTestCollection.json";
+import WhatsAppBusinessTestCollection from "../../assets/WhatsAppBusinessTestCollection.json";
```
g10z3r, line 7: Let’s add a comment here like: `// FIXME: remove when ...`

g10z3r (approved):

jevgenis33 (approved):

view/desktop/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,87 @@
...
+        const node = findNodeById(activeTree, pathNode)!;
+        const lastItem = index === path.length - 1;
+
+        if (lastItem) {
+          return (
+            <div key={pathNode} className="contents">
+              <TestCollectionIcon type={node.type} className="size-4.5" />
+              <span>{pathNode}</span>
+            </div>
+          );
```
g10z3r: Would it make sense to extract this into a separate function like `const activeNode = useMemo(() => ... )`?

view/desktop/src/components/Tree/utils.ts
```diff
@@ -67,10 +68,27 @@ export const removeUniqueIdFromTree = (tree: TreeNodeProps): NodeProps => {
   };
 };
 
+export const findNodeById = (tree: NodeProps, id: string): NodeProps | undefined => {
+  if (tree.id === id) {
+    return tree;
+  }
+  for (const child of tree.childNodes) {
```
g10z3r, line 75: I would add a check if (tree.childNodes && tree.childNodes.length > 0) to ensure that tree.childNodes exists and is not null or undefined before iterating over it

view/desktop/src/components/Tree/utils.ts
```diff
@@ -67,10 +68,27 @@ export const removeUniqueIdFromTree = (tree: TreeNodeProps): NodeProps => {
...
+  }
+
+  return undefined;
+};
+
 export const findNodeByUniqueId = (tree: TreeNodeProps, uniqueId: string): TreeNodeProps | undefined => {
   if (tree.uniqueId === uniqueId) return tree;
 
-  return tree.childNodes.find((child) => findNodeByUniqueId(child, uniqueId));
+  for (const child of tree.childNodes) {
```
g10z3r, line 86: Same here (if (tree.childNodes && tree.childNodes.length > 0) )

view/desktop/src/components/Breadcrumbs/Breadcrumbs.tsx
Enthceph: no, Breadcrumbs component renders only once

PR #87, opened 2025-04-08
refactor(SAPIC-141): [Backend] Implementation of DB manager reload functionality

brutusyhy:


crates/moss-collection/src/collection.rs
```diff
@@ -80,14 +81,12 @@ impl Collection {
...
         })
     }
 
-    pub fn state_db_manager(&self) -> Result<Arc<dyn StateDbManager>> {
-        self.state_db_manager
-            .clone()
-            .ok_or(anyhow!("The state_db_manager has been dropped"))
+    pub fn state_db_manager(&self) -> Arc<dyn StateDbManager> {
+        self.state_db_manager.clone()
     }
```
g10z3r, line 90: I don't think this function makes sense anymore

crates/moss-collection/src/collection.rs
```diff
@@ -204,21 +203,13 @@ impl Collection {
...
-        self.state_db_manager = Some(Arc::new(state_db_manager_impl));
-
-        Ok(())
+    pub async fn reset(&mut self, new_path: &Path) -> Result<()> {
+        if let Some(manager) = Arc::get_mut(&mut self.state_db_manager) {
+            self.abs_path = new_path.to_path_buf();
+            manager.reset(self.fs.clone(), new_path).await
+        } else {
+            Err(anyhow!("The Collection StateDbManager is being used"))
+        }
```
g10z3r, line 212: This should be an internal implementation within the DB manager, the collection should simply call the reload method on the DB manager. And I’m not sure this can actually work, we always have a reference wrapped in `Arc`, so this will always cause a panic?

crates/moss-collection/src/storage/state_db_manager.rs
```diff
@@ -2,13 +2,17 @@ use super::request_store::TABLE_REQUESTS;
...
 use std::sync::Arc;
+use async_trait::async_trait;
+use moss_fs::{FileSystem, RenameOptions};
 
 const COLLECTION_STATE_DB_NAME: &str = "state.db";
 
 pub struct StateDbManagerImpl {
-    request_store: Arc<dyn RequestStore>,
+    path: PathBuf,
+    // FIXME: It appears that we need to wrap it in Option for temporary dropping in safe Rust
```
g10z3r, line 14: Unnecessary comment

crates/moss-collection/src/storage/state_db_manager.rs
```diff
@@ -18,12 +22,28 @@ impl StateDbManagerImpl {
...
+    async fn reset(&mut self, fs: Arc<dyn FileSystem>, new_path: &Path) -> Result<()> {
+        std::mem::take(&mut self.request_store);
+
+        let old_path = std::mem::replace(&mut self.path, new_path.to_path_buf());
+        fs.rename(&old_path, &new_path, RenameOptions::default()).await?;
+
+        let db_client = ReDbClient::new(&self.path.join(COLLECTION_STATE_DB_NAME))?
+            .with_table(&TABLE_REQUESTS)?;
+        self.request_store = Some(Arc::new(RequestStoreImpl::new(db_client)));
+        Ok(())
```
g10z3r, line 45: Hmm... There should be a single place where we define the tables... right now, we’re defining them both here and in the `new` function

brutusyhy: Maybe we'll extract a utility method here?

PR #90, opened 2025-04-09
devops(SAPIC-142): [DevOps] Python dependencies file

brutusyhy:
Fix the script to be case sensitive, add a `requirements.txt` that lists the necessary dependencies for python scripts, and update the `make ready` command to install the dependencies


Makefile
```diff
@@ -23,13 +23,15 @@ DESKTOP_DIR := view/desktop
 ICONS_DIR := tools/icongen
 # --- Tool Directories ---
 XTASK_DIR := tools/xtask
+PYTHON_SCRIPTS_DIR := misc
```
g10z3r, line 26: I’m not sure about that — for now, let’s just use the path directly in the command, since we don’t need this variable anywhere else.

brutusyhy: Ok

PR #92, opened 2025-04-11
feat(SAPIC-144): [Backend] Integration tests for the Request specific functionality

brutusyhy:


crates/moss-fs/src/utils.rs
```diff
@@ -30,6 +32,36 @@ pub fn decode_directory_name(encoded: &str) -> Result<String, std::num::ParseInt
 }
 
 
+
+// FIXME: This process may need some refinement
+/// Doing a basic normalization using Path::components() and encode the segments after the prefix
+pub fn encode_path(prefix: Option<&Path>, path: &Path) -> Result<PathBuf> {
```
g10z3r, line 38: Let’s order the arguments by importance, first the path, then the optional prefix.

crates/moss-fs/src/utils.rs
```diff
@@ -30,6 +32,36 @@ pub fn decode_directory_name(encoded: &str) -> Result<String, std::num::ParseInt
...
+        path.strip_prefix(prefix)?
+    } else { path };
+
+    let encoded_part = path_to_be_encoded.components().filter_map(|c| {
+        match c {
+            Component::Normal(os_str) => {
+                let segment = os_str.to_string_lossy();
+                Some(encode_directory_name(&segment))
+            },
+            // FIXME: Is this the best strategy?
```
g10z3r, line 49: I think this code can be rewritten in a more idiomatic Rust style (the comments in the code are for clarity and don’t need to be included in the codebase).

```rust
pub fn encode_path(prefix: Option<&Path>, path: &Path) -> Result<PathBuf> {
    // Determine the relative part of the path to be encoded.
    let relative_path = match prefix {
        Some(prefix) => path.strip_prefix(prefix)?,
        None => path,
    };

    // Encode only the normal components of the path.
    let encoded: PathBuf = relative_path
        .components()
        .filter_map(|comp| {
            if let Component::Normal(name) = comp {
                Some(encode_directory_name(&name.to_string_lossy()))
            } else {
                // Special components are ignored (ParentDir, Prefix, RootDir, CurDir)
                None
            }
        })
        .collect();

    // If a prefix was provided, join it back with the encoded path.
    Ok(prefix.map(|p| p.join(&encoded)).unwrap_or(encoded))
}

```

crates/moss-fs/src/utils.rs
```diff
@@ -1,3 +1,5 @@
+use anyhow::Result;
+use std::path::{Component, Path, PathBuf};
 use regex::Regex;
```
g10z3r, line 4: The `encode_directory_name` function can also be optimized — we can move the regular expression outside the function and make it precompiled.

```rust 
static FORBIDDEN_RE: Lazy<Regex> = Lazy::new(|| {
    Regex::new(r#"[.%<>:"/\\|?*]"#).expect("...")
});

...
```

g10z3r (approved):

PR #103, opened 2025-04-17
feat(SAPIC-146): [Backend] Integration tests for the DB crate

brutusyhy:
1. Remove `config` parameter from encrypt and decrypt methods
2. Compact newly created databases


crates/moss-db/src/lib.rs
g10z3r, line 1: I’d also consider adding tests for the `with_table` method. Maybe they’re already there and I just didn’t notice, but if not, I think it’s worth adding them

brutusyhy: Good idea!

g10z3r (approved):

PR #105, opened 2025-04-18
feat(SAPIC-153): [Backend] Solution for the data serialization issue

brutusyhy:
`bincode` serializes the data in a non self-describing way: this means that it cannot be used to serialize complex Rust enums with recursion. `serde_json` will serialize the metadata alongside with the data, thus it can handle such types with no issue. We are still storing binary representation in the database, which means that there should only be a modest increase in data size, reflecting the overhead from JSON format, compared to `bincode`.
I have also written a test for writing complex data types into a `BincodeTable` and reading from it.

### DIFF
--- a/crates/moss-db/src/bincode_table.rs
+++ b/crates/moss-db/src/bincode_table.rs
@@ -65,7 +65,6 @@ where
             Transaction::Write(txn) => {
                 let mut table = txn.open_table(self.table)?;
 
-                // Serialize the value as JsonValue and save it as bytes
                 let bytes = serde_json::to_vec(value)?;
 
                 table.insert(key.borrow(), bytes)?;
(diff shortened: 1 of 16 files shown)


crates/moss-db/src/bincode_table.rs
```diff
@@ -63,11 +64,14 @@ where
         match txn {
             Transaction::Write(txn) => {
                 let mut table = txn.open_table(self.table)?;
-                let bytes = bincode::serialize(value)?;
+
+                // Serialize the value as JsonValue and save it as bytes
```
g10z3r, line 68: I think the comment here is unnecessary the code is pretty clear and doesn’t need extra explanation

crates/moss-db/src/bincode_table.rs
```diff
@@ -77,17 +81,18 @@ where
         match txn {
             Transaction::Write(txn) => {
                 let mut table = txn.open_table(self.table)?;
-                let value = table
+
+                // Get the bytes of JsonValue and deserialize it
```
g10z3r, line 85: I think the comment here is unnecessary. The same applies to all other comments in similar places.

PR #106, opened 2025-04-19
feat(SAPIC-167): Forwarding of the Workspace API to the frontend and implementation of missing operations

g10z3r:


crates/moss-workspace/tests/shared/mod.rs
```diff
@@ -28,19 +33,40 @@ pub fn random_workspace_path() -> PathBuf {
...
+
+    let app_dir = random_app_dir_path.clone();
+    let cleanup_fn = Box::new(move || {
+        let path = app_dir;
+        Box::pin(async move {
+            if let Err(e) = tokio::fs::remove_dir_all(&path).await {
+                eprintln!("Failed to clean up test directory: {}", e);
+            }
+        }) as Pin<Box<dyn Future<Output = ()> + Send>>
+    });
```
g10z3r, line 67: @brutusyhy, now we have resources required for the workspace manager to function, but we don’t want to expose them in tests, I added a new return value: a cleanup callback that's created in this function and simply called inside test functions

crates/moss-workspace/bindings/types.ts
```diff
@@ -53,4 +53,4 @@ export type PanelRenderer = "onlyWhenVisible" | "always";
 
 export type SidebarPartState = { preferredSize: number; isVisible: boolean };
 
-export type WorkspaceInfo = { path: string; name: string };
+export type WorkspaceInfo = { path: string; name: string; lastOpenedAt?: bigint };
```
g10z3r, line 56: @Enthceph, I added a new field called `lastOpenedAt`, which you can use to implement the “Recent” list on the Welcome page. Basically, you fetch all existing workspaces and sort them by this field.

PR #107, opened 2025-04-19
feat(SAPIC-166): [Backend] Implementation of all missing functionality in collection operations and test fixes

brutusyhy:
1. Update all the request group operations to also update the RequestNode map and state db.
2. Introduce a new error check for passing an incorrect type of resource (using a request group key in a request operation, for example)
3. Update the request group tests to also test that the RequestNode map is updated


crates/moss-collection/src/collection/api/create_request_group.rs
```diff
@@ -1,38 +1,71 @@
...
+                    .unwrap_or_default()
+                    .to_string_lossy()
+                    .to_string(),
+                path: encoded_path,
+                order: None,
+                spec_file_name: None,
+            }))
+        };
+
+        dbg!(request_store.scan()?);
```
g10z3r, line 65: forget to remove that? 😄

crates/moss-collection/src/collection/api/create_request_group.rs
```diff
@@ -1,38 +1,71 @@
...
+        input: CreateRequestGroupInput,
     ) -> Result<CreateRequestGroupOutput, OperationError> {
         input.validate()?;
 
         let encoded_path = encode_path(&input.path, None)?;
-        let request_group_full_path = self
-            .abs_path
-            .join(REQUESTS_DIR)
-            .join(&encoded_path);
+        let request_group_full_path = self.abs_path.join(REQUESTS_DIR).join(&encoded_path);
```
g10z3r, line 19: If it’s an absolute path, maybe it makes sense to name it `abs_path`?

crates/moss-collection/src/collection/api/rename_request.rs
```diff
@@ -10,11 +10,15 @@ impl Collection {
...
 
-        let requests = self.registry().await?.requests_nodes();
-        let mut requests_lock = requests.write().await;
+        let request_nodes = self.registry().await?.requests_nodes();
+        let mut requests_lock = request_nodes.write().await;
 
         let mut lease_request_data = requests_lock.lease(input.key)?;
 
+        if !lease_request_data.is_request() {
+            return Err(anyhow!("Resource {} is not a request", input.key).into());
```
g10z3r, line 19: Shouldn’t this be considered a validation error, or..?

crates/moss-collection/src/collection/api/rename_request_group.rs
```diff
@@ -52,38 +52,51 @@ impl Collection {
...
         let new_encoded_name = encode_name(&input.new_name);
-        let group_relative_path_new = group_relative_path_old
+        let group_dir_relative_path_new = group_dir_relative_path_old
             .parent()
             .unwrap()
             .join(&new_encoded_name);
-
-        if group_relative_path_old == group_relative_path_new {
+        dbg!(&group_dir_relative_path_old);
+        dbg!(&group_dir_relative_path_new);
```
g10z3r, line 75: Forgotten debug statements in the code 😄

crates/moss-collection/src/collection/api/rename_request_group.rs
```diff
@@ -52,38 +52,51 @@ impl Collection {
...
+            }
+
+            group_data.path().to_owned()
+        };
+
         let new_encoded_name = encode_name(&input.new_name);
-        let group_relative_path_new = group_relative_path_old
+        let group_dir_relative_path_new = group_dir_relative_path_old
             .parent()
             .unwrap()
```
g10z3r, line 72: Is this safe? I’d avoid using `unwrap` in internal modules, better to handle errors more gracefully there

crates/moss-collection/src/collection/api/rename_request_group.rs
```diff
@@ -52,38 +52,51 @@ impl Collection {
...
-            .join(&group_relative_path_old);
+            .join(&group_dir_relative_path_old);
 
-        if !group_full_path_old.exists() {
+        if !group_dir_full_path_old.exists() {
             return Err(OperationError::NotFound {
-                name: group_relative_path_old
+                name: group_dir_relative_path_old
                     .file_name()
                     .unwrap_or_default()
```
g10z3r, line 89: If we hit the `unwrap_or_default`, what exactly will be substituted in?

crates/moss-collection/src/collection/error.rs
```diff
@@ -19,6 +19,7 @@ pub enum OperationError {
 
     #[error("unknown error: {0}")]
     Unknown(#[from] anyhow::Error),
+    // FIXME: Should we have an error for incorrect entity type?
```
g10z3r, line 22: I think this could be treated as validation errors

crates/moss-collection/src/collection/api/rename_request.rs
brutusyhy: I guess that makes sense

crates/moss-collection/src/collection/api/rename_request_group.rs
brutusyhy: This is actually safe: parent() will only return a None when the path terminates in a root, a prefix or is empty. This should never happen here

crates/moss-collection/src/collection/api/rename_request_group.rs
brutusyhy: Here, I also think it would be impossible for the unwrap to fail, as the paths we generate will never end in "..", which is how `file_name()` can fail

g10z3r (approved):

PR #119, opened 2025-04-25
feat(SAPIC-157): Frontend Integration of Workspace CRUD API operations into the UI

Enthceph:


view/desktop/src/pages/WelcomePage/WelcomePageDivider.tsx
g10z3r, line 1: Does it make sense to include the word “Page” in the filenames stored in the `/pages/Welcome` folder? It seems a bit redundant, since they’re already under the `pages` directory.

view/desktop/src/pages/WelcomePage/WelcomePageRecentWorkspaces.tsx
```diff
@@ -0,0 +1,43 @@
+import { useEffect, useState } from "react";
+
+import { Button } from "@/components";
+import { useGetWorkspaces } from "@/hooks/useGetWorkspaces";
+import { useOpenWorkspace } from "@/hooks/useOpenWorkspace";
+import { useWorkspaceStore } from "@/store/workspace";
+
+import WelcomePageLink from "./WelcomePageLink";
+
+export const WelcomePageRecentWorkspaces = () => {
```
g10z3r, line 10: These components will never be used outside of this page, so simple names like `Recent`, `PinBoard`, etc., should be perfectly sufficient

view/desktop/src/pages/WelcomePage/WelcomePageDivider.tsx
Enthceph: Well, I (and if i'm not mistaken Jevgenis) initially said the same thing, but you sad that it's better to always include a name prefix for page components to avoid confusion—especially in cases where we have generic component names like Settings or Home. That’s why I added the WelcomePage prefix to all relevant components. This ensures that when using VSCode's IntelliSense (e.g., with Ctrl+. to import), we don't accidentally import a component with name Link that's meant exclusively for the WelcomePage.

Enthceph: I can remove it though, it's not that hard

g10z3r (approved):

PR #136, opened 2025-05-02
devops(SAPIC-186): Implementation of a utility to generate React components from SVG icons

brutusyhy:
Implement a python script that generates React components for svg icons. Run `make new-gen-icons` to test the script.


misc/svg_component_generator.py
```diff
@@ -0,0 +1,294 @@
...
+    icon_plans = {}
+    if plan_file.exists():
+        icon_plans = json.loads(plan_file.read_text())
+
+    plan_updated = False
+    for icon_name in icon_list:
+        if icon_name in icon_plans and not force:
+            continue
+        print(f"Updating the plan for {icon_name}")
+        icon_plans[icon_name] = plan_icon(folder, icon_name)
```
g10z3r, line 113: I think instead of using `print`, it's better to use the standard `logging` module and add a `--verbose` flag to control the output level

Example
```python 
def setup_logging(verbose: bool = False) -> None:
    """
    Configure logging for the module.
    """
    level = logging.DEBUG if verbose else logging.INFO
    logging.basicConfig(
        format="[%(asctime)s] %(levelname)s: %(message)s",
        level=level,
        datefmt="%Y-%m-%d %H:%M:%S",
    )


```

misc/svg_component_generator.py
```diff
@@ -0,0 +1,294 @@
...
+
+            new_class_name = f"{prev_class_name} {class_name}".strip()
+            result_elem.set("className", new_class_name)
+
+    return result
+
+
+
+if __name__ == "__main__":
+    ET.register_namespace("", "http://www.w3.org/2000/svg")
```
g10z3r, line 274: That `http://www.w3.org/2000/svg` string could be extracted into a constant

misc/svg_component_generator.py
```diff
@@ -0,0 +1,294 @@
...
+    icon_list = sorted(icon_name for icon_name in os.listdir(folder) if not Path(folder, icon_name).is_file())
+
+    # Check if there's an update to the icon list
+    hash_string = "".join(f"{len(name)}{name}" for name in icon_list)
+    icon_hash = hashlib.md5(hash_string.encode()).hexdigest()
+    lock_file = Path(folder, "plan.lock")
+    if lock_file.exists() and lock_file.read_text() == icon_hash and not force:
+        print("No changes to the icon list")
+        return
+    lock_file.write_text(icon_hash)
```
g10z3r, line 100: We should only update the lock file after successful icon generation. Otherwise, if the generation fails, we’ll have an updated hash and the next planning step won’t be triggered even though the icons weren’t actually processed

Makefile
```diff
@@ -79,7 +79,17 @@ ready: gen-icons
 ## Generate application icons
 .PHONY: gen-icons
 gen-icons:
-	@cd $(ICONS_DIR) && $(PNPM) build --silent 
+	@cd $(ICONS_DIR) && $(PNPM) build --silent
+
+## New icon generator tool
+.PHONY: new-gen-icons
+new-gen-icons:
+	@cd $(MISC_DIR) && $(PIP) install --break-system-packages -r requirements.txt
```
g10z3r, line 87: Installation shouldn’t be part of the command itself. We already have a `make ready` command for installing all dependencies

jevgenis33 (approved):

g10z3r (approved):

PR #147, opened 2025-05-08
devops(SAPIC-223): Replacing the Python scripts with a TypeScript utility

brutusyhy:
Replace `ts_imports_consolidator.py` and `ts_imports_injector.py` with two simple TS scripts that use `ts-morph`, and reorganize the tool structure by placing the TS scripts in a package under the `tools` folder


jevgenis33 (approved):
Great work!

PR #148, opened 2025-05-08
feat(SAPIC-225): Improve workspace management

jevgenis33:


view/desktop/src/components/Modals/Workspace/NewWorkspaceModal.tsx
```diff
@@ -64,7 +64,7 @@ export const NewWorkspaceModal = ({ closeModal, showModal }: ModalWrapperProps)
               value={name}
               className="max-w-72"
               onChange={(e) => setName(e.target.value)}
-              pattern={'^[^\\/:\\*\\?"><>|]+$'}
+              pattern={'^[^/:*?<>|"\\\\]+$'}
```
g10z3r, line 67: There shouldn’t be a regex here. For now, the only requirement is that the name must have a minimum length of 1

view/desktop/src/hooks/appState/useWorkspaceState.ts
```diff
@@ -0,0 +1,33 @@
+import { useDescribeAppState } from "./useDescribeAppState";
+
+export type WorkspaceStateType = "empty" | "inWorkspace";
```
g10z3r, line 3: I think it would be better to represent this as an enum

```typescript
export const enum WorkbenchState {
	EMPTY = 1,
	WORKSPACE
}
```

And this should be located somewhere else. The hook file should only export the hook itself and its mutation key constant, if there is one.

PR #154, opened 2025-05-14
feat(SAPIC-227): Implementation of the remaining operations for collection requests and writing the tests

brutusyhy:
Implement the remaining operations for collection requests and tests for the operations.


crates/moss-collection/tests/collection__create_request_dir_entry.rs
```diff
@@ -0,0 +1,99 @@
...
+    assert!(changed_paths.iter().any(|(path, _id, kind)| {
+        path.to_path_buf() == PathBuf::from("requests") && kind == &PathChangeKind::Created
+    }));
+    assert!(changed_paths.iter().any(|(path, _id, kind)| {
+        path.to_path_buf() == PathBuf::from("requests").join(&request_dir_name)
+            && kind == &PathChangeKind::Created
+    }));
+
+    // Clean up
+    { tokio::fs::remove_dir_all(&collection_path).await.unwrap() }
```
g10z3r, line 36: image

That kind of code structure will trigger a warning notification during compilation.
It should be done like this instead

```rust
{
...
}
```

There’s also a plan to replace this with a cleanup function, similar to how it’s currently implemented in the `workspace/workbench` crate. 

So you can either just tweak it to avoid the warnings, or go ahead and implement the cleanup function right away and replace those spots with it.

crates/moss-collection/tests/collection__create_request_dir_entry.rs
g10z3r, line 1: The `__` underscore is no longer needed in file names (see examples in `workspace/workbench`)

crates/moss-collection/src/collection/api.rs
```diff
@@ -1,13 +1,9 @@
-pub mod create_request_group;
-pub mod create_request_old;
-pub mod delete_request;
-pub mod delete_request_group;
 pub mod list_requests;
```
g10z3r, line 1: That’s also a "obsolete" API operation

crates/moss-collection/src/collection/api/delete_request_entry.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        let entry = {
+            let snapshot_lock = worktree.read().await;
+            let entry = snapshot_lock
+                .entry_by_id(input.id)
+                .ok_or(OperationError::NotFound {
+                    name: input.id.to_string(),
+                    path: Default::default(),
+                })?;
+            // TODO: Check unit type?
```
g10z3r, line 25: No, that’s not needed here. I’m also considering removing this field from the `Entry` structure entirely, since it’s more like frontend-related data.

crates/moss-collection/src/models/operations.rs
g10z3r, line 1: Can we delete the code that’s currently commented out, or do we still need it?

crates/moss-collection/src/models/operations.rs
```diff
@@ -184,19 +215,26 @@ pub struct UpdateRequestDirEntryOutput {
...
-pub struct DeleteRequestDirEntryInput {
+pub struct UpdateRequestEntryInput {
     pub id: EntryId,
+
+    /// A new name for the request, if provided,
+    /// the request will be renamed to this name.
+    #[ts(optional)]
+    #[validate(length(min = 1))]
+    pub name: Option<String>,
+    // TODO: url and payload?
```
g10z3r, line 231: Not yet, for now, we’re only updating the name.

crates/moss-collection/src/models/operations.rs
```diff
@@ -255,3 +293,73 @@ fn validate_request_destination(destination: &Path) -> Result<(), ValidationErro
...
+
+    if let Some(extension) = destination.extension() {
+        if &extension.to_string_lossy().to_string() == ".request" {
+            return Err(ValidationError::new(
+                "Request dir name should not end in '.request'",
+            ));
+        }
+    }
+    Ok(())
+}
```
g10z3r, line 312: I forgot to do that in my previous PR but the idea was that we’d have a more general function that takes in the expected extension (`.request`, `.endpoint`, etc.), and then the specific validation functions would use that. Let’s go ahead and implement that now, so it’s easier to extend going forward.

crates/moss-testutils/src/random_name.rs
```diff
@@ -12,8 +12,8 @@ pub fn random_request_name() -> String {
     format!("Test_{}_Request", random_string(10))
 }
 
-pub fn random_request_group_name() -> String {
-    format!("Test_{}_Group", random_string(10))
+pub fn random_request_dir_name() -> String {
+    format!("Test_{}_Dir", random_string(10))
```
g10z3r, line 16: This doesn’t seem like the best place for a general utility function  it’s specific to `moss-collection`, so I think it’d be better to place it there.

crates/moss-collection/tests/collection__create_request_dir_entry.rs
brutusyhy: I see

crates/moss-collection/src/models/operations.rs
brutusyhy: I think we can delete them

g10z3r (approved):

PR #156, opened 2025-05-14
devops(SAPIC-241): Cleanup in the tools folder

brutusyhy:
Move ts scripts related to bindings generation to tools/gen-bindings


pnpm-workspace.yaml
```diff
@@ -2,5 +2,5 @@ packages:
   - "crates/*"
   - "view/*"
   - "packages/*"
-  - "tools"
+  - "tools/gen-bindings"
```
g10z3r, line 5: Should it definitely be `"tools/gen-bindings"` and not `"tools/*"`?

g10z3r (approved):

pnpm-workspace.yaml
brutusyhy: Let me test it

PR #159, opened 2025-05-16
refactor(SAPIC-244): Implementation and integration of the SanitizedName type

brutusyhy:


crates/moss-common/src/sanitized.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+pub struct SanitizedName(String);
+impl SanitizedName {
+    pub fn new(raw: &str) -> Self {
+        Self(encode(raw))
+    }
+
+    pub fn to_original(&self) -> String {
+        decode(&self.0).expect("SanitizedName should be correctly encoded")
```
g10z3r, line 45: We shouldn’t use `unwrap` here, we don’t want to crash the entire app just because we failed to decode a string.

g10z3r, line 45: Also, I’d consider caching the original value, that way we wouldn’t have to perform the decoding operation every time.

brutusyhy: Ok

g10z3r (approved):

PR #162, opened 2025-05-17
feat(SAPIC-253): Standardization of the application’s storages

g10z3r:


crates/moss-storage/src/storage/operations.rs
```diff
@@ -57,6 +57,7 @@ pub trait PutItem: Send + Sync {
     fn put(&self, key: Self::Key, entity: Self::Entity) -> DatabaseResult<()>;
 }
 
+// TODO: Should remove return the value?
```
g10z3r, line 60: I think yes, if we’re able to do it

PR #164, opened 2025-05-17
feat(SAPIC-185): Add an animated background to the new collection creation area.

SMikulins:


view/desktop/src/components/CollectionTreeView/CollectionTreeView.css
g10z3r, line 1: Try to solve this without using CSS. Also, whenever you’re using a color, remember that it should depend on the currently selected theme.

PR #165, opened 2025-05-18
feat(SAPIC-254):  Implementation of the frontend representation of worktree entries

g10z3r:


crates/moss-collection/src/api/stream_entries_by_prefixes.rs
```diff
@@ -27,34 +29,41 @@ impl Collection {
...
 
         let worktree = self.worktree().await?;
-        let snapshot_lock = worktree.snapshot().await.read().await;
-
-        if snapshot_lock.count_files() == 0 {
-            // We need to send a final empty event to signal the end of the stream.
-            let _ = channel.send(StreamEntriesByPrefixesEvent(vec![]));
-            return Ok(());
-        }
+        let read_lock = worktree.read().await;
```
g10z3r, line 32: It’s better to name things like this more descriptively, like `worktree_lock` so that it’s clear later in the code what exactly is being locked

crates/moss-collection/src/api/stream_entries_by_prefixes.rs
```diff
@@ -27,34 +29,41 @@ impl Collection {
...
+        //
+        // if snapshot_lock.count_files() == 0 {
+        //     // We need to send a final empty event to signal the end of the stream.
+        //     let _ = channel.send(StreamEntriesByPrefixesEvent(vec![]));
+        //     return Ok(());
+        // }
```
g10z3r, line 38: I assume we still need this, otherwise, the frontend won’t get the signal that the stream has been stopped in this case

crates/moss-collection/src/models/operations.rs
```diff
@@ -15,131 +20,76 @@ use crate::models::{
...
     pub id: EntryId,
 }
 
 #[derive(Clone, Debug, Serialize, TS)]
 #[serde(rename_all = "camelCase")]
 #[ts(export, export_to = "operations.ts")]
-pub struct DeleteRequestDirEntryOutput {
-    pub changed_paths: Arc<[(Arc<Path>, EntryId, PathChangeKind)]>,
+pub struct DeleteEntryOutput {
+    pub physical_changes: ChangesDiffSet,
```
g10z3r, line 58: The information from `physical_changes` shouldn’t be exposed to the frontend, and this field shouldn’t be present in the TypeScript structure. For the most part, this data is here for debugging and testing purposes.

g10z3r: The same applies to other structures that send diffs to the frontend

crates/moss-collection/src/models/operations.rs
```diff
@@ -15,131 +14,80 @@ use crate::models::{
 /// All the path and file names passed in the input should be unencoded.
 /// For example, a name of "workspace.name" will be encoded as "workspace%2Ename"
 /// The frontend should simply use the name and path used in the user's original input
```
g10z3r, line 16: this I think is no longer relevant

crates/moss-collection/src/worktree/virtual_worktree.rs
```diff
@@ -0,0 +1,155 @@
...
+};
+use std::path::PathBuf;
+use std::{
+    path::Path,
+    sync::{Arc, atomic::AtomicUsize},
+};
+
+pub struct VirtualWorktree {
+    next_entry_id: Arc<AtomicUsize>,
+    // FIXME: Do we need a lock here?
```
g10z3r, line 18: No, the lock should be acquired at the level of the entire Worktree to ensure consistency of changes

crates/moss-collection/src/models/operations.rs
brutusyhy: I see! How do we handle parts of tests that check if the physical changes are correct?

crates/moss-collection/src/worktree/virtual_worktree.rs
g10z3r, line 1: I think in this file we don’t need to check `debug_assert!(old_path.is_relative());` since the paths are entirely virtual

crates/moss-collection/src/worktree/virtual_worktree.rs
brutusyhy: Got it

crates/moss-collection/src/worktree/virtual_worktree.rs
brutusyhy: That makes sense

crates/moss-collection/src/api/stream_entries_by_prefixes.rs
g10z3r: Yes, but the idea here is to avoid going deeper into the function and instead exit early if the tree is empty, using a "early return" pattern. We can check the number of virtual entries, and if it's 0, just return from the function right away.

brutusyhy: I see

PR #185, opened 2025-06-01
feat(SAPIC-291): Implement `AnyValue` as an enum

brutusyhy:


crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+    fn test_anyvalue_enum() -> Result<()> {
+        let db_path = PathBuf::from("tests").join("anyvalue_enum.db");
+        let table: BincodeTable<i32, AnyValueEnum> = BincodeTable::new("any_value");
+        let client = ReDbClient::new(&db_path)?.with_table(&table)?;
+        let test_data = TestData {
+            name: "secret".to_string(),
+            id: 42,
+        };
+
+        {
```
g10z3r, line 148: In this case, you don’t need to return the result from the test. All places that could produce an error should be unwrapped with `assert` so that debugging is quicker and easier if something goes wrong.

If the result isn’t important (for example, we’re not testing the transaction launch itself), then it’s fine to just call `unwrap` right away.

crates/moss-db/src/anyvalue_enum.rs
g10z3r, line 1: Creating a whole separate module(file) for this seems a bit excessive

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+            AnyValueEnum::Text(_) => "Text",
+            AnyValueEnum::Bytes(_) => "Bytes",
+        }
+    }
+    pub fn is_null(&self) -> bool {
+        match self {
+            AnyValueEnum::Null => true,
+            _ => false,
+        }
+    }
```
g10z3r, line 34: JUst matches!(self, AnyValueEnum::Null)

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+        match self {
+            AnyValueEnum::Int(i) => Some(*i),
+            _ => None,
+        }
+    }
+
+    pub fn as_unsized_int(&self) -> Option<u64> {
+        match self {
+            AnyValueEnum::UnsizedInt(u) => Some(*u),
+            _ => None,
```
g10z3r, line 53: Should we also add   AnyValueEnum::Int(i) if *i >= 0 => Some(*i as u64) ?

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
+    pub fn as_double(&self) -> Option<f64> {
+        match self {
+            AnyValueEnum::Double(d) => Some(*d),
+            _ => None,
```
g10z3r, line 60: AnyValueEnum::Int(i) => Some(*i as f64),
AnyValueEnum::UnsizedInt(u) => Some(*u as f64),

?

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
```
g10z3r, line 63: `as_bytes` is not presented 

Something like 

    pub fn as_bytes(&self) -> Option<&[u8]> {
        match self {
            AnyValueEnum::Bytes(b) => Some(b.as_slice()),
            _ => None,
        }
    }
    
    Would also add 
    
       pub fn try_as_number<T>(&self) -> Option<T>
    where
        T: TryFrom<i64> + TryFrom<u64> + TryFrom<f64>,
    {
        match self {
            AnyValueEnum::Int(i) => T::try_from(*i).ok(),
            AnyValueEnum::UnsizedInt(u) => T::try_from(*u).ok(),
            AnyValueEnum::Double(d) => T::try_from(*d).ok(),
            _ => None,
        }
    }

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+        match self {
+            AnyValueEnum::Text(t) => Some(t.as_str()),
+            _ => None,
+        }
+    }
+
+    pub fn serialize<T: Serialize>(value: &T) -> Result<Self> {
+        serde_json::to_vec(value)
+            .map(|bytes| Self::Bytes(bytes))
+            .map_err(|e| e.into())
```
g10z3r, line 74: Lets map it with more context 

 .map_err(|e| anyhow!("Serialization failed: {}", e))

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    pub fn deserialize<T: DeserializeOwned>(&self) -> Result<T> {
+        match self {
+            AnyValueEnum::Bytes(bytes) => serde_json::from_slice(bytes).map_err(|e| e.into()),
+            _ => Err(anyhow!("cannot deserialize non-bytes value")),
```
g10z3r, line 80: Maybe it make we can add a type name? 

Err(anyhow!(
                "Cannot deserialize non-bytes value: expected Bytes variant, got {}",
                self.type_name()
            ))

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
+use anyhow::{Result, anyhow};
+use redb::{Key, TypeName, Value};
+use serde::{Deserialize, Serialize, de::DeserializeOwned};
+use std::cmp::Ordering;
+
+#[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
+pub enum AnyValueEnum {
```
g10z3r, line 7: We need to implement PartialOrd and Ord 

   fn type_order(&self) -> u8 {
        match self {
            AnyValueEnum::Null => 0,
            AnyValueEnum::Bool(_) => 1,
            AnyValueEnum::Int(_) => 2,
            AnyValueEnum::UnsizedInt(_) => 3,
            AnyValueEnum::Double(_) => 4,
            AnyValueEnum::Text(_) => 5,
            AnyValueEnum::Bytes(_) => 6,
        }
    }
    
    impl PartialOrd for AnyValueEnum {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}

impl Ord for AnyValueEnum {
    fn cmp(&self, other: &Self) -> Ordering {
        // First compare by type
        match self.type_order().cmp(&other.type_order()) {
            Ordering::Equal => {
                // If same type, compare values
                match (self, other) {
                    (AnyValueEnum::Null, AnyValueEnum::Null) => Ordering::Equal,
                    (AnyValueEnum::Bool(a), AnyValueEnum::Bool(b)) => a.cmp(b),
                    (AnyValueEnum::Int(a), AnyValueEnum::Int(b)) => a.cmp(b),
                    (AnyValueEnum::UnsizedInt(a), AnyValueEnum::UnsizedInt(b)) => a.cmp(b),
                    (AnyValueEnum::Double(a), AnyValueEnum::Double(b)) => Self::compare_f64(*a, *b),
                    (AnyValueEnum::Text(a), AnyValueEnum::Text(b)) => a.cmp(b),
                    (AnyValueEnum::Bytes(a), AnyValueEnum::Bytes(b)) => a.cmp(b),
                    _ => unreachable!("Type order should prevent this case"),
                }
            }
            other_ordering => other_ordering,
        }
    }
}

g10z3r, line 7: This should implement 

 From<bool> 
 From<i64> 
 From<i32>
 From<u64>
 From<u32>
  From<f64> 
  From<f32> 
  From<String>
  From<&str> 
  From<&[u8]> 
  From<Vec<u8>>

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    fn fixed_width() -> Option<usize> {
+        None
+    }
+
+    fn from_bytes<'a>(data: &'a [u8]) -> Self::SelfType<'a>
+    where
+        Self: 'a,
+    {
+        // FIXME: Is this always true?
```
g10z3r, line 103: If JSON deserialization fails, store as raw bytes
 
         serde_json::from_slice(data).unwrap_or_else(|_| {
            AnyValueEnum::Bytes(data.to_vec())
        })

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+        serde_json::from_slice(data)
+            .expect("Should be able to deserialize bytes back to AnyValueEnum")
+    }
+
+    fn as_bytes<'a, 'b: 'a>(value: &'a Self::SelfType<'b>) -> Self::AsBytes<'a>
+    where
+        Self: 'b,
+    {
+        // FIXME: Is this always true?
```
g10z3r, line 112: Fallback to empty vec if serialization fails

    serde_json::to_vec(value).unwrap_or_else(|_| {
            vec![]
        })

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    fn type_name() -> TypeName {
+        TypeName::new("AnyValueEnum")
+    }
+}
+
+impl Key for AnyValueEnum {
+    fn compare(data1: &[u8], data2: &[u8]) -> Ordering {
+        std::cmp::Ord::cmp(data1, data2)
```
g10z3r, line 123: Deserialize and compare logically instead of byte comparison

Something like 
       let val1 = Self::from_bytes(data1);
        let val2 = Self::from_bytes(data2);
        val1.cmp(&val2)

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
+use anyhow::{Result, anyhow};
+use redb::{Key, TypeName, Value};
+use serde::{Deserialize, Serialize, de::DeserializeOwned};
+use std::cmp::Ordering;
+
```
g10z3r, line 5: Might be also usefull to have such macro 

#[macro_export]
macro_rules! any_value {
    (null) => {
        $crate::anyvalue_enum::AnyValueEnum::Null
    };
    ($value:expr) => {
        $crate::anyvalue_enum::AnyValueEnum::from($value)
    };
}

crates/moss-db/src/anyvalue_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+impl Key for AnyValueEnum {
+    fn compare(data1: &[u8], data2: &[u8]) -> Ordering {
+        std::cmp::Ord::cmp(data1, data2)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
g10z3r, line 128: We need to test type_conversions, ordering, nan_handling and macro as well

crates/moss-db/src/anyvalue_enum.rs
brutusyhy: It was just so that I can write `?` instead of calling unwrap() lol

crates/moss-db/src/anyvalue_enum.rs
brutusyhy: Yes, I was creating a new file since I didn't want to change anything existing at this point.

crates/moss-db/src/anyvalue_enum.rs
brutusyhy: That makes sense!

PR #199, opened 2025-06-06
feat(SAPIC-296) Implementation of the log deletion feature

brutusyhy:
Overhauled how rolling logs work by replacing `tracing-appender` with our custom writer implementation: it will write up to `dump_threshold` number of log entries in an in-memory buffer, and dump them into files from time to time. Since the writer never has to keep holding a file handle, there's no problem with having queries and updates.

Also introduced two api `delete_log` and `delete_logs`. While technically we only need ID to identify a log entry, a timestamp helps us to narrow down our scope of search.


crates/moss-logging/src/api/delete_log.rs
```diff
@@ -0,0 +1,113 @@
...
+    io::Write,
+    path::{Path, PathBuf},
+};
+
+use crate::{
+    FILE_DATE_FORMAT, LoggingService, TIMESTAMP_FORMAT,
+    models::{operations::DeleteLogInput, types::LogEntry},
+};
+impl LoggingService {
+    pub fn delete_log(&self, input: &DeleteLogInput) -> Result<()> {
```
g10z3r, line 14: If this is an operation, then the return type should be `OperationResult`

crates/moss-logging/src/api/delete_log.rs
```diff
@@ -0,0 +1,113 @@
...
+            let idx = sessionlog_queue_lock.iter().position(|x| x.id == input.id);
+            if let Some(idx) = idx {
+                sessionlog_queue_lock.remove(idx);
+                return Ok(());
+            }
+        }
+        let datetime = DateTime::parse_from_str(&input.timestamp, TIMESTAMP_FORMAT)?;
+        // 2. Check if the log is inside app logs and update it
+        {
+            let log_files = Self::identify_log_file(&self.applog_path, datetime)?;
```
g10z3r, line 35: There’s no need to number the steps. If we ever need to rearrange or add new ones, we’d have to update comments in code that ideally shouldn’t be affected by those changes.

crates/moss-logging/src/api/delete_log.rs
```diff
@@ -0,0 +1,113 @@
...
+            for file in log_files {
+                let updated = Self::update_log_file(&file, &input.id)?;
+                if updated {
+                    return Ok(());
+                }
+            }
+        }
+
+        // FIXME: Result for non-existent log entry?
+        Ok(())
```
g10z3r, line 55: Any API endpoint that performs a deletion should return a structure containing the ID of the record that was deleted.

g10z3r: Moreover, I’d include a field in the response with the path of the file from which the record was deleted. This should be used in tests and debugging, but it shouldn’t be part of the exported TypeScript structure.

crates/moss-logging/src/api/delete_log.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl LoggingService {
+    fn identify_log_file(path: &Path, datetime: DateTime<FixedOffset>) -> Result<Vec<PathBuf>> {
+        // Use timestamp to identify which file to update
+        // We just need to check the last two log files that starts before the log entry's timestamp
+        // Two because of potential timestamp rounding issue
+
+        // OPTIMIZE: We might use a binary search here but I'm not sure if it's necessary
+        let mut file_list = Vec::new();
+        for entry in std::fs::read_dir(path)? {
```
g10z3r, line 67: Everywhere, you should use an instance of `Arc<dyn FS>`

g10z3r, line 67: I think we should keep an in-memory list of all the files we’re interested in. That way, we won’t need to read the directory every time

crates/moss-logging/src/api/delete_log.rs
```diff
@@ -0,0 +1,113 @@
...
+            let entry = entry?;
+            let path = entry.path();
+            if path.is_file() {
```
g10z3r, line 70: In cases like this, it’s better to use the [early return pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8) to reduce code nesting

crates/moss-logging/src/api/delete_log.rs
```diff
@@ -0,0 +1,113 @@
...
+            .into_iter()
+            .filter(|item| item.1 <= datetime.naive_local())
+            .rev()
+            .take(2)
+            .map(|item| path.join(item.0).with_extension("log"))
+            .collect::<Vec<_>>();
+        Ok(files)
+    }
+
+    fn update_log_file(path: &Path, id: &str) -> Result<bool> {
```
g10z3r, line 89: This function needs a few improvements:

* Use  self.fs to save memory when working with large files.
* Only write back to the file if something was actually removed.
* Eliminate unnecessary allocations.
* Return a more informative error.

crates/moss-logging/src/models/operations.rs
```diff
@@ -18,3 +19,18 @@ pub struct ListLogsInput {
...
+pub struct DeleteLogInput {
+    pub timestamp: String, // This helps us to narrow down the scope of search
+    pub id: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export, export_to = "operations.ts")]
+pub struct DeleteLogsInput {
+    pub inputs: Vec<DeleteLogInput>,
```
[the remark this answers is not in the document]
brutusyhy: The list of log entries that we want to delete

crates/moss-logging/src/models/operations.rs
```diff
@@ -18,3 +19,18 @@ pub struct ListLogsInput {
 pub struct ListLogsOutput {
     pub contents: Vec<LogEntry>,
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export, export_to = "operations.ts")]
+pub struct DeleteLogInput {
+    pub timestamp: String, // This helps us to narrow down the scope of search
```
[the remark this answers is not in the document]
brutusyhy: No. This is the timestamp string we have in the log entry, such as "2025-06-05T13:15:59.017+0300"

crates/moss-logging/src/api/delete_log.rs
brutusyhy: At least for RollingWriter I don't think we can use that: the tracing_subscriber only takes in a synce std::io::Write object, so we shouldn't use our async filesystem there.

brutusyhy: But for each timestamp, we look for different files. For example, if you provide a log entry with a timestamp of 15:00:30, we want to look at file for 14:59 and 15:00 (since the latter might begin later than 15:00:30, and this entry is actually in 14:59 file). The files we want to look at would change for different timestamps

g10z3r (approved):

PR #202, opened 2025-06-07
feat(SAPIC-284): Improve state management

jevgenis33:


view/desktop/src/components/ActivityBar.tsx
```diff
@@ -49,27 +49,29 @@ export const ActivityBar = () => {
...
-        "w-full border-t border-t-(--moss-border-color) px-1.5": position === "bottom",
-        "h-full flex-col py-1.5": position === "default",
-        "hidden": position === "hidden",
+        "w-full border-b border-b-(--moss-border-color) px-1.5": position === "TOP",
+        "w-full border-t border-t-(--moss-border-color) px-1.5": position === "BOTTOM",
+        "h-full flex-col py-1.5": position === "DEFAULT",
+        "hidden": position === "HIDDEN",
 
-        "border-l border-l-(--moss-border-color)": sideBarPosition === "right" && position === "default",
+        "border-l border-l-(--moss-border-color)": sideBarPosition === "RIGHT" && position === "DEFAULT",
```
g10z3r, line 57: I think it's better to use the [primitive types](https://github.com/moss-foundation/sapic/blob/main/crates/moss-workspace/bindings/primitives.ts) that's exported from the backend here. 

And the same applies to all other similar places

view/desktop/src/hooks/workspace/useGlobalSidebarState.ts
```diff
@@ -0,0 +1,45 @@
+import { useEffect, useRef } from "react";
+import { useAppResizableLayoutStore } from "@/store/appResizableLayout";
+import { useActiveWorkspace } from "./useActiveWorkspace";
+import { SidebarPosition } from "@repo/moss-workspace";
+
+const GLOBAL_SIDEBAR_DEFAULTS = {
+  width: 255,
+  visible: true,
+  position: "LEFT" as SidebarPosition,
+};
```
g10z3r, line 10: The frontend shouldn't define its own defaults

PR #209, opened 2025-06-12
refactor(SAPIC-298): Introducing the AnyValue enum into the workspace and workbench crates

brutusyhy:
Replace the use of `AnyValue` with `AnyValueEnum`. Right now, we are still using `serialize` method for all values, but in the future we will only use `serialize` on object values; for primitive values, we will simply use `any_value!` macro.


g10z3r:
bugbot run

PR #219, opened 2025-06-16
refactor(SAPIC-327) Dependency cleanup

brutusyhy:
Remove less frequently used crates such as `parking_lot`, `arc-swap` and `dashmap`, several other unused dependencies, and temporarily disable git crates/dependencies, since on Windows clearing cargo cache would require rebuilding openssl2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,8 +102,8 @@ jobs:
       fail-fast: false
       matrix:
         package:
-          # - moss_git
-          # - moss_git_hosting_provider
+          - moss_git
+          - moss_git_hosting_provider
           - moss_tauri
           - moss_text
           - moss_collection
@@ -161,8 +161,8 @@ jobs:
       fail-fast: false
       matrix:
         package:
-          # - moss_git
-          # - moss_git_hosting_provider
+          - moss_git
+          - moss_git_hosting_provider
           - moss_tauri
           - moss_text
           - moss_collection
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,8 @@
+## Note for development on Windows
+
+`libgit2` has a known bug on Windows which uses a faulty libssh2 backend, which won't be able to recognize new key
+formats (https://github.com/rust-lang/git2-rs/issues/1139)
+To make it work, we will need to compile a compatible version of openssl during build.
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -31,8 +31,8 @@ moss_collection = { path = "crates/moss-collection" }
 moss_db = { path = "crates/moss-db" }
 moss_app = { path = "crates/moss-app" }
 moss_applib = { path = "crates/moss-applib" }
-moss_git = { path = "crates/moss-git" }
-moss_git_hosting_provider = { path = "crates/moss-git-hosting-provider" }
+# moss_git = { path = "crates/moss-git" }
```
g10z3r, line 34: No, we shouldn’t do that. We shouldn’t exclude working crates just because we’re not actively using them

crates/moss-app/src/services/log_service.rs
```diff
@@ -184,7 +183,7 @@ impl LogService {
                 OperationError::InvalidInput("The input timestamp is invalid".to_string())
             })?;
         {
-            let mut applog_queue_lock = self.applog_queue.lock();
+            let mut applog_queue_lock = self.applog_queue.lock().expect("Mutex poisoned");
```
g10z3r, line 186: Why do we need a panic here?

crates/moss-app/src/services/log_service/rollinglog_writer.rs
```diff
@@ -31,7 +35,7 @@ impl<'a> std::io::Write for RollingLogWriter {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
         let log_entry: LogEntryInfo = serde_json::from_str(String::from_utf8_lossy(buf).as_ref())?;
 
-        let mut queue_lock = self.log_queue.lock();
+        let mut queue_lock = self.log_queue.lock().expect("Mutex poisoned");
```
g10z3r, line 38: Why do we need a panic here?

crates/moss-app/src/services/log_service.rs
brutusyhy: As I mentioned to you last time, std Mutex has a poison mechanism: a lock is considered poisoned when any thread holding that lock panics, leaving the data in an undetermined state. Essentially, this only panics when we already have another panics somewhere else

PR #222, opened 2025-06-17
feat(SAPIC-324): Implemented a universal method for bulk log deletion and refactored the service

brutusyhy:
1. Generalize the `batch_delete_log` method and simplify its logic by storing which file a log entry is written to.
2. Implement `LogServiceError` and `LogServiceResult`
3. Migrate the tests

### DIFF
--- a/crates/moss-db/src/common.rs
+++ b/crates/moss-db/src/common.rs
@@ -1,5 +1,5 @@
 use redb::{ReadTransaction as InnerReadTransaction, WriteTransaction as InnerWriteTransaction};
-use std::io::{Error, ErrorKind};
+use std::io::ErrorKind;
 use thiserror::Error;
 
 pub type AnyEntity = Vec<u8>;


crates/moss-app/src/api/list_logs.rs
```diff
@@ -0,0 +1,32 @@
...
+use moss_common::api::OperationResult;
+use tauri::Runtime as TauriRuntime;
+
+use crate::{
+    app::App,
+    models::operations::{ListLogsInput, ListLogsOutput},
+    services::log_service::{LogFilter, LogService},
+};
+
+// TODO: impl App
```
g10z3r, line 11: What’s the purpose of this comment? 😄

crates/moss-app/src/services/log_service/rollinglog_writer.rs
```diff
@@ -1,32 +1,53 @@
...
         }
     }
 }
 
+fn map_database_error_to_io_error(error: DatabaseError) -> std::io::Error {
+    std::io::Error::new(
+        ErrorKind::Other,
+        format!("Database operation failed: {}", error),
+    )
+}
```
g10z3r, line 49: In Rust, unlike other languages, we can implement elegant From/Into traits instead of functions like these

```rust
impl Into<std::io::Error> for DatabaseError {
    fn into(self) -> std::io::Error {
        std::io::Error::new(std::io::ErrorKind::Other, self.to_string())
    }
}
```

crates/moss-app/src/services/log_service/rollinglog_writer.rs
```diff
@@ -37,23 +58,41 @@ impl<'a> std::io::Write for RollingLogWriter {
...
+                let file_path = self.log_path.join(file_name).with_extension("log");
 
                 let file = OpenOptions::new()
                     .create(true)
                     .append(true)
-                    .open(&self.log_path.join(file_name).with_extension("log"))?;
+                    .open(&file_path)?;
                 let mut writer = BufWriter::new(file);
+
+                let mut write_txn = self
```
g10z3r, line 70: In our case, the transaction can only be write, so the write prefix seems redundant

PR #226, opened 2025-06-18
feat(SAPIC-295): Refinement of the welcome page

SMikulins:


Enthceph:
We use our [Scrollbar component](https://github.com/moss-foundation/sapic/blob/main/view/desktop/src/lib/ui/Scrollbar.tsx) to make scrollbar styled instead of default OS styles
изображение

g10z3r (approved):

PR #228, opened 2025-06-20
feat(SAPIC-268): Added the ability to set a collection repo and upload an avatar image

brutusyhy:
Add the ability to upload avatar images for collections, as well as setting and updating a collection's repo path


crates/moss-collection/src/collection.rs
```diff
@@ -54,10 +61,20 @@ pub struct CreateParams<'a> {
     pub name: Option<String>,
     pub internal_abs_path: &'a Path,
     pub external_abs_path: Option<&'a Path>,
+    pub repo: Option<Url>,
+    pub icon_path: Option<PathBuf>,
+}
+
+pub enum IconModification {
+    Nochange,
```
g10z3r, line 69: That’s redundant. Just wrap this type in an Option so the field logic stays consistent everywhere. If the field isn’t present, it means no updates are needed for the corresponding data

crates/moss-collection/src/collection.rs
```diff
@@ -141,6 +160,15 @@ impl Collection {
         )
         .await?;
 
+        // If the user provides an icon, transform it and save it to assets/icon.png
```
g10z3r, line 163: The comment is redundant, the code it explains is already pretty simple and clear

crates/moss-collection/src/lib.rs
```diff
@@ -23,4 +24,8 @@ pub mod dirs {
     pub const COMPONENTS_DIR: &str = "components";
     pub const SCHEMAS_DIR: &str = "schemas";
     pub const ENVIRONMENTS_DIR: &str = "environments";
+    pub const ASSETS_DIR: &str = "assets";
 }
+
+pub const ICON_NAME: &str = "icon.png";
+pub const ICON_SIZE: u32 = 128;
```
g10z3r, line 31: Group it all and keep it inside `pub mod constants {}`

crates/moss-collection/src/manifest.rs
```diff
@@ -2,26 +2,36 @@ use anyhow::Result;
...
 use serde::{Deserialize, Serialize};
 use toml_edit::DocumentMut;
+use url::Url;
 
 pub(crate) const MANIFEST_FILE_NAME: &str = "Sapic.toml";
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ManifestModel {
     pub name: String,
+    // TODO: Validation of repo path?
```
g10z3r, line 12: Validation should be at the DTO level, not at the manifest model level

crates/moss-collection/src/manifest.rs
```diff
@@ -2,26 +2,36 @@ use anyhow::Result;
...
+    // We can have two types of repo paths though: HTTPS and SSH
```
g10z3r, line 13: This can only be a link like `github.com/moss-foundation/sapic` only HTTP, and without explicitly specifying the protocol

crates/moss-collection/src/manifest.rs
```diff
@@ -2,26 +2,36 @@ use anyhow::Result;
...
+    pub repo: Option<Url>,
 }
 
 #[derive(Debug)]
 pub struct ManifestModelDiff {
     /// A new name for the collection, if provided, the collection
     /// will be renamed to this name.
     pub name: Option<String>,
+    /// A new repo link for the collection, if provided, the collection
+    /// will be relinked to this repo.
```
g10z3r, line 23: This can only be a link like github.com/moss-foundation/sapic only HTTP, and without explicitly specifying the protocol

crates/moss-collection/src/manifest.rs
g10z3r, line 1: Let’s probably use the full word `repository` instead of the shortened `repo`, since we’re not abbreviating the other fields either

crates/moss-collection/src/services/image_upload.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::Result;
+use image::{GenericImageView, imageops::FilterType};
+use std::path::Path;
+
+pub struct ImageUploadService {}
+
+impl ImageUploadService {
+    pub fn upload_icon(img_path: &Path, output_path: &Path, icon_size: u32) -> Result<()> {
```
g10z3r, line 8: It’s probably more like `set_icon` than `upload`, since from the backend’s perspective we’re not actually uploading anything from anywhere

crates/moss-collection/src/services/image_upload.rs
```diff
@@ -0,0 +1,26 @@
...
+        // Crop to centered square
+        let side = w.min(h);
+        let x0 = (w - side) / 2;
+        let y0 = (h - side) / 2;
+        let square = image::imageops::crop_imm(&img, x0, y0, side, side).to_image();
+
+        // Resize down/up to your avatar_size
+        let icon = image::imageops::resize(&square, icon_size, icon_size, FilterType::Lanczos3);
+
+        // Save to the given output path
```
g10z3r, line 22: [Redundant comments](https://levelup.gitconnected.com/stop-writing-bad-comments-in-your-code-tips-for-clean-code-73f1af5eaf31)

crates/moss-workspace/src/models/operations.rs
```diff
@@ -25,28 +25,48 @@ pub struct CreateCollectionInput {
 
     #[ts(optional)]
     pub external_path: Option<PathBuf>,
+
+    #[ts(optional)]
+    pub repo: Option<Url>,
+
+    #[ts(optional)]
+    pub icon_path: Option<PathBuf>,
```
g10z3r, line 33: Let’s use the new directive on the entire struct for optional fields

crates/moss-collection/src/collection.rs
```diff
@@ -141,6 +160,15 @@ impl Collection {
...
+        if let Some(icon_path) = params.icon_path {
+            ImageUploadService::upload_icon(
+                &icon_path,
+                &abs_path.join(ASSETS_DIR).join(ICON_NAME),
+                ICON_SIZE,
+            )?;
```
g10z3r, line 169: It’s better to just log the error here since a missing image doesn’t affect the functionality of the collection

g10z3r (approved):

PR #232, opened 2025-06-23
feat(SAPIC-351): Canonicalization of the Git Url

brutusyhy:


crates/moss-git/src/url.rs
```diff
@@ -0,0 +1,28 @@
+use anyhow::{Result, anyhow};
+use url::Url;
+
+// Strip the protocol and ".git" suffix from a git repo url
+pub fn clean_git_url(repo_url: &Url) -> Result<String> {
```
g10z3r, line 5: I think this is more like normalize\_git\_url than clean\_git\_url

g10z3r, line 5: This should be integrated into the process of writing to the manifest file

PR #239, opened 2025-06-28
feat(SAPIC-343): Forwarding WASM functions to HCL context 

brutusyhy:
Implement `WasmHost` for loading and executing WASM plugins, and provide examples of writing WASM plugins in JavaScript, Python and Rust (misc/wasm_plugins). 
Also include an example of using WASM plugins during evaluation of HCL.


Cargo.toml
```diff
@@ -20,8 +20,10 @@ members = [
     "crates/moss-activity-indicator",
     "crates/moss-file",
     "crates/moss-hcl",
+    "crates/moss-wasm",
     "tools/xtask",
     "view/desktop/bin",
+    "misc/wasm_plugins/rust_demo", "misc/wasm_plugins/hcl_greeter", "misc/wasm_plugins/echo",
```
g10z3r, line 26: I think it’s better to create these as separate packages that don’t depend on the workspace, so the workspace manifest file doesn’t get cluttered

g10z3r (approved):

PR #279, opened 2025-07-14
refactor(SAPIC-395): [Backend] First experience using joinerror

brutusyhy:


crates/moss-api/src/lib.rs
```diff
@@ -24,6 +24,13 @@ pub enum TauriError {
     Timeout,
 }
 
+// FIXME: How do we best convert joinerror::Error to TauriError?
```
g10z3r, line 27: I think that’s the right approach. We'll specifically keep a separate TauriError type to clearly separate domain/infrastructure errors from the handler layer

crates/moss-workspace/tests/workspace__update_collection.rs
```diff
@@ -97,11 +97,8 @@ async fn rename_collection_empty_name() {
         )
         .await;
 
-    assert!(matches!(
-        rename_collection_result,
-        Err(OperationError::InvalidInput(_))
-    ));
-
+    dbg!(&rename_collection_result);
```
g10z3r, line 100: Forgotten debug statement

g10z3r (approved):

PR #285, opened 2025-07-17
feat(SAPIC-404): Implement the main concepts of `AtomicFileSystem`

brutusyhy:


Cargo.toml
```diff
@@ -37,7 +37,7 @@ members = [
     #
     # Tools
     #
-    "tools/xtask",
+    "tools/xtask", "crates/moss-atomic-fs",
```
g10z3r, line 40: This should be a library, not an internal crate

crates/moss-atomic-fs/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+impl Drop for FileSystemTransaction {
+    fn drop(&mut self) {
+        // Clean up the temp_folder for the transaction
+        let _ = std::fs::remove_dir_all(&self.temp_folder);
+    }
+}
+
+// TODO: Allow only one transaction at a time?
+
+pub struct AtomicFileSystem {}
```
g10z3r, line 69: AtomicFileSystem is what will be in our moss-fs crate it’s a feature of how we implement working with the file system, but it shouldn’t be part of the library itself. 

Let’s drop the functionality for creating without rollback for now—we can add that later at the moss-fs level. The library itself should give us functionality that we can use like this:

```rust
use atomic_fs::{self, Rollback};

let rollback = Rollback::new();
atomic::create_dir(rollback, "test", Options::default());
atomic::create_file(rollback, "test/test.txt", Options::default());

rollback.apply();
```

crates/moss-atomic-fs/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+    },
+    RemoveFile(PathBuf),
+    Rename {
+        from: PathBuf,
+        to: PathBuf,
+        options: RenameOptions,
+    },
+}
+pub struct FileSystemTransaction {
+    temp_folder: PathBuf,
```
g10z3r, line 56: It’s better to stick to the standard and shorten "temporary" to "tmp"

crates/moss-atomic-fs/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+        content: Vec<u8>,
+        options: CreateOptions,
+    },
+    RemoveFile(PathBuf),
+    Rename {
+        from: PathBuf,
+        to: PathBuf,
+        options: RenameOptions,
+    },
+}
```
g10z3r, line 54: Instead of a list of actions, could we just work with a list of callbacks?

crates/moss-atomic-fs/src/lib.rs
g10z3r, line 1: Before finishing the task, I think we should remove all obvious comments and resolve all the TODOs

g10z3r (approved):

PR #308, opened 2025-08-06
feat(SAPIC-432): Integrating storage into environment operations

brutusyhy:


crates/moss-workspace/src/storage/entities/state_store.rs
```diff
@@ -45,3 +45,9 @@ pub struct EditorPanelStateEntity {
     pub maximum_width: Option<f64>,
     pub maximum_height: Option<f64>,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct EnvironmentEntity {
```
g10z3r, line 50: Same as with variables. If this data can be updated separately, then it shouldn’t be in a single structure.

Updates should be by keys:

environment:<id>:order

And I guess, we should store this in the resources table

g10z3r (approved):

PR #322, opened 2025-08-14
feat(SAPIC-459): Returning information about the current branch

brutusyhy:
A. In `StreamCollectionEvent` we will also return information about the current branch:

1. Branch name
2. Remote state (counts of ahead and behind commits)

B. Introduce `Workspace::describe_collection` endpoint which can be called when the collection page is open. We will provide repository and contributors info here instead of during `stream_collections`

C. Update the logic of getting branch_info so that fetch failure will not fail the operation (it will return the un-updated value), and make ahead and behind commits optional so that at least the current branch name is returned.

D. Fix unable to delete `.git` folder after cloning a repo by recursively set all git objects as writable and drop the repo_handle before deleting the folder.

E. Allow workspace to be opened even if it contains invalid collections.

TODO:
Local state (whether there’s anything to commit): To be done after clarifying the logic of staging files and making commits


crates/moss-git/src/repo.rs
```diff
@@ -225,6 +226,19 @@ impl RepoHandle {
             .collect::<Vec<_>>();
         Ok(names)
     }
+
+    pub fn current_branch(&self) -> joinerror::Result<String> {
+        let head = self.repo.head()?;
+
+        // If HEAD points to a branch, get the branch name.
+        if head.is_branch() {
+            let branch_name = head.shorthand().unwrap_or("unknown branch");
```
g10z3r, line 235: If the branch is not found, I think we should return None without using magic values. Deciding what to return to the frontend if the branch is not found should not be handled at this crate’s level

crates/moss-workspace/src/models/events.rs
```diff
@@ -20,6 +20,7 @@ pub struct StreamCollectionsEvent {
 
     pub repository_info: Option<RepositoryInfo>,
```
g10z3r, line 21: We can remove repository_info, contributors from here; they will only be in CollectionInfo.

crates/moss-workspace/src/models/events.rs
```diff
@@ -20,6 +20,7 @@ pub struct StreamCollectionsEvent {
...
     pub contributors: Vec<Contributor>,
+    pub branch_info: Option<BranchInfo>,
```
g10z3r, line 23: I think just the name “branch” will be enough

crates/moss-workspace/src/api/stream_collections.rs
```diff
@@ -62,26 +75,65 @@ impl<R: AppRuntime> Workspace<R> {
...
+
+async fn fetch_current_branch_info(
+    repo_handle: Arc<Mutex<Option<RepoHandle>>>,
+) -> joinerror::Result<Option<BranchInfo>> {
+    let result = tokio::task::spawn_blocking(move || {
+        let repo_handle_lock = repo_handle.lock()?;
+        let repo_handle_ref = repo_handle_lock.as_ref();
+        if repo_handle_ref.is_none() {
+            return Ok(None);
+        }
```
g10z3r, line 114: I would pass a mutable reference to the repository here right away. That way, you wouldn’t need to check for its existence in the function, and the return type would be simpler

crates/moss-workspace/src/api/stream_collections.rs
```diff
@@ -62,26 +75,65 @@ impl<R: AppRuntime> Workspace<R> {
...
-                .await
-                .unwrap_or_default(),
-        ),
-        _ => (None, Vec::new()),
+    if let Some(Ok(repo_ref)) = repository.as_ref().map(|x| GitUrl::parse(&x)) {
+        match repo_ref.domain.as_str() {
+            // FIXME: Handle custom GitLab domains
+            "github.com" => (
+                github_client.repository_info(&repo_ref).await.ok(),
+                github_client
```
g10z3r, line 87: It doesnt seem like the best code organization… going into a function that’s supposed to perform an action and passing None into it. Call the function that does the specific action only if you have a repository. Check for its presence at a higher level

g10z3r (approved):

PR #348, opened 2025-08-31
devops(SAPIC-500): Macro for exporting constants to type script

brutusyhy:
Introduce `#[const_export]` macro for exporting Rust constants to TypeScript. At this moment, it will be exported to `constants.ts` in the same location as `index.ts`, which is the same as what we are doing manually right now. The binding generation scripts are also updated so that the exports for constants are correctly injected.


crates/moss-bindingutils/const-export/Cargo.toml
g10z3r, line 1: It's better to give the folder a more general name, like in the example, since we might add new macros in the future.

crates/moss-app/src/lib.rs
```diff
@@ -36,17 +36,9 @@ impl<R: AppRuntime> ActiveWorkspace<R> {
...
-    // ###                                                            ###
-    // ### !!! PLEASE UPDATE THE TYPESCRIPT CONSTANTS IN index.ts !!! ###
-    // ###                                                            ###
-    // ##################################################################
-
-    // When adding/removing/modifying any constants here,
-    // you must also update the corresponding TypeScript constants
-    // in moss-app/index.ts to match your changes.
-
+    #[const_export]
```
g10z3r, line 41: There should be a way to specify exactly where the constant is being exported to otherwise, it feels like magic

PR #349, opened 2025-09-01
devops(SAPIC-352): Extend cargo with a flag for creating a crate with bindings

brutusyhy:
Introduce `cargo-tsnew` to quickly create a crate with typescript bindings. Run `make ready` first to install the tool, and use it this way:
"cargo tsnew [crate_name]"
This will do the normal `cargo new --lib [crate_name]`, as well as generating `index.ts`, `package.json` and `tsconfig.json`


tools/cargo-tsnew/src/main.rs
g10z3r, line 1: For consistency with `cargo new`, I’d name it `new-ts` to make it more convenient and intuitive to use

tools/cargo-tsnew/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    "skipLibCheck": true,
+    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "allowSyntheticDefaultImports": true,
+    "strict": true,
+    "noEmit": true
+  },
+  "include": ["**/*.ts"],
+  "exclude": ["node_modules"]
```
g10z3r, line 54: Does it make sense to do this using the `json!` macro?

tools/cargo-tsnew/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    if args.is_empty() || args[0] == "--help" || args[0] == "-h" {
+        println!("cargo tsnew <name>");
+        return;
+    }
+
+    // `cargo new {crate_name} --lib
+    let crate_name = &args[0];
+    let mut cargo_new = Command::new("cargo");
+    cargo_new.arg("new").arg("--lib").arg(crate_name);
```
g10z3r, line 71: I think the `lib` flag shouldn’t be added automatically. If it wasn’t specified, then we don’t include it

tools/cargo-tsnew/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    let crate_dir = Path::new(crate_name);
+
+    // Create `index.ts`, `package.json` and `tsconfig.json`
+    fs::write(crate_dir.join("index.ts"), index_ts_content()).unwrap();
+    fs::write(
+        crate_dir.join("package.json"),
+        package_json_content(crate_name),
+    )
+    .unwrap();
+    fs::write(crate_dir.join("tsconfig.json"), tsconfig_json_content()).unwrap();
```
g10z3r, line 88: It's better to move the predefined files into a global/local unified hashmap like `<filename>:<Option<content>>` so everything is gathered in one place

PR #362, opened 2025-09-04
feat(SAPIC-499) Implementation of Collection Archive Export/Import Feature

brutusyhy:
Implement collection archive file export and import feature. Exporting a collection will bundle every file/directory (except for local files such as config, state databases and .git folder) into a zip file, and importing an archive file will extract those files and create a collection based on it.


crates/moss-collection/src/collection.rs
```diff
@@ -32,8 +32,12 @@ use crate::{
     services::{set_icon_service::SetIconService, storage_service::StorageService},
     vcs::{CollectionVcs, Vcs},
     worktree::Worktree,
+    zip::zip_dir,
 };
 
+const ARCHIVE_EXCLUDED_DIRECTORIES: [&'static str; 1] = [".git"];
```
g10z3r, line 38: Let’s combine this into a single list of paths to exclude.

crates/moss-collection/src/collection.rs
```diff
@@ -439,8 +443,28 @@ impl<R: AppRuntime> Collection<R> {
 
         Ok(())
     }
-}
 
+    pub async fn export_archive(&self, out_file: &Path) -> joinerror::Result<()> {
```
g10z3r, line 447: I suppose we should also check that the destination path is actually a folder?

crates/moss-collection/src/zip.rs
```diff
@@ -0,0 +1,120 @@
...
+        return Err(Error::new::<()>(
+            "cannot export archive file into the source folder",
+        ));
+    }
+
+    // Wrap synchronous zip operations so that it doesn't block async runtime
+    let src_dir = src_dir.to_path_buf();
+    let out_file = out_file.to_path_buf();
+
+    tokio::task::spawn_blocking(move || {
```
g10z3r, line 26: Why is this needed here?

crates/moss-workspace/bindings/types.ts
```diff
@@ -116,6 +121,14 @@ export type EnvironmentInfo = {
   variables: Array<VariableInfo>;
 };
 
+export type ExportCollectionParams = {
+  id: string;
+  /**
+   * Path to the output zip file
+   */
+  outFile: string;
```
g10z3r, line 129: Is this a path or a file?

crates/moss-workspace/bindings/types.ts
```diff
@@ -177,7 +190,10 @@ export type ImportCollectionParams = {
 /**
  * @category Type
  */
-export type ImportCollectionSource = { "gitHub": GitHubImportParams } | { "gitLab": GitLabImportParams };
+export type ImportCollectionSource =
+  | { "gitHub": GitHubImportParams }
+  | { "gitLab": GitLabImportParams }
+  | { "archiveFile": ArchiveFileImportParams };
```
g10z3r, line 196: Let’s give this a more concise name - archive.

crates/moss-workspace/src/services/collection_service.rs
```diff
@@ -622,6 +630,108 @@ impl<R: AppRuntime> CollectionService<R> {
...
+            icon_path,
+            abs_path,
+            external_path: None,
+            archived: false,
+        })
+    }
+
+    pub(crate) async fn export_collection(
+        &self,
+        _ctx: &R::AsyncContext,
```
g10z3r, line 721: Do you need a context here ?

g10z3r (approved):

PR #377, opened 2025-09-13
feat(SAPIC-461) Workspace vcs api

brutusyhy:
Introduce basic vcs apis for collection, including commit/push/discard/fetch/pull. The endpoint is `execute_vcs_operation`. Right now the changes that are committed/discarded are passed as relative paths, but in the future we will encapsulate them so that the frontend works on the basis of entries, rather than raw files.


crates/moss-collection/src/vcs.rs
```diff
@@ -127,11 +132,60 @@ impl<R: AppRuntime> CollectionVcs<R> for Vcs<R> {
...
+
+        Ok(())
+    }
+
+    // Pushing currently checked-out branch to the configured refspec+remote
+    async fn push<'a>(&self, ctx: &R::AsyncContext) -> joinerror::Result<()> {
+        let repo_lock = self.repository.read().await;
+        let repo_ref = repo_lock
+            .as_ref()
+            .ok_or_join_err::<()>("repository handle is dropped")?;
```
g10z3r, line 167: Lets mark this errors as an Internal

crates/moss-collection/src/worktree.rs
```diff
@@ -808,9 +808,7 @@ async fn process_entry(
             },
             desc,
         )));
-    }
-
-    if item_config_path.exists() {
+    } else if item_config_path.exists() {
```
g10z3r, line 811: Can we use the early return pattern here?

g10z3r (approved):

PR #386, opened 2025-09-23
feat(SAPIC-538): Implement validation of the JSON representation of the theme using a Rego policy

brutusyhy:


policies/theme.rego
```diff
@@ -0,0 +1,121 @@
...
+}
+
+color_ok(color) := true if {
+	color.type == "gradient"
+    gradient_ok(color.value)
+}
+
+color_ok(color) := true if {
+	color.type == "variable"
+    # TODO: Variable validation rule?
```
g10z3r, line 43: Yeah, I think it could be useful

crates/moss-app/src/theme/install.rs
```diff
@@ -0,0 +1,29 @@
...
+pub fn install_theme(
+    input_path: &Path,
+    policy_path: &Path,
+    themes_dir: &Path,
+) -> joinerror::Result<()> {
+    let theme: Theme = serde_json::from_reader(std::fs::File::open(input_path)?)?;
+    validate_theme(policy_path, &theme)?;
+
+    let css = convert_theme_to_css(&theme)?;
+    // TODO: should the filename be based on identifier?
```
g10z3r, line 18: For now, the name should remain so that backward compatibility is preserved. When there's a full installation mechanism, the installation process itself will be identical to the process in VS Code.

policies/theme.rego
g10z3r, line 1: I think tests for the main points will be useful

crates/moss-app/src/theme/install.rs
brutusyhy: Ok

PR #387, opened 2025-09-24
feat(SAPIC-541): Capabilities for creating and managing external projects

brutusyhy:
Implement creating project at an external path and restoring projects from external path

### DIFF
--- a/crates/moss-workspace/src/models/types.rs
+++ b/crates/moss-workspace/src/models/types.rs
@@ -246,10 +246,10 @@ pub struct EditorPartStateInfo {
 #[serde(rename_all = "camelCase")]
 #[ts(export, export_to = "types.ts")]
 pub enum ImportProjectSource {
-    GitHub(GitHubImportParams),
-    GitLab(GitLabImportParams),
-    Archive(ArchiveImportParams),
-    Disk(DiskImportParams),
+    GitHub(ImportGitHubParams),
+    GitLab(ImportGitLabParams),
+    Archive(ImportArchiveParams),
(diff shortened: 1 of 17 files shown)


crates/moss-workspace/src/models/types.rs
```diff
@@ -289,6 +289,14 @@ pub struct ArchiveImportParams {
     pub archive_path: PathBuf,
 }
 
+/// @category Type
+#[derive(Debug, Serialize, Deserialize, TS, Validate)]
+#[serde(rename_all = "camelCase")]
+#[ts(export, export_to = "types.ts")]
+pub struct ExternalImportParams {
```
g10z3r: I guess, the name here should be something like ImportFromDiskParams. And I think all the other types should be renamed so that the word order in the type name is correct: {Action}{Domain}{Type Specifics}.
 
ImportArchiveParams
ImportGitLabParams 

and so on

crates/moss-workspace/src/project.rs
```diff
@@ -767,12 +773,124 @@ impl<R: AppRuntime> ProjectService<R> {
...
+                .join_err::<()>("failed to start transaction")?;
+
+            self.storage
+                .put_item_order_txn(ctx, &mut txn, id.as_str(), params.order)
+                .await?;
+            self.storage
+                .put_expanded_items_txn(ctx, &mut txn, &state_lock.expanded_items)
+                .await?;
+
+            txn.commit()?;
```
g10z3r, line 872: I think this TODO can be handled right away. Just log the non-critical errors and that's it.

